PR #2, opened 2017-03-31
user_guide: clarify endian override for floats

davidhicks:
Make it clearer in the user guide that float types (f4/f8) can have
their endian nature overridden by specifying a suffix of either be or
le. For example, allowable types are f4 and f8 for the default endian
setting, and f4be, f4le, f8be, f8le for explicitly specified endian
float values.


GreyCat:
Thanks!

PR #3, opened 2017-04-01
Update documentation to accurately reflect latest available meta fields

davidhicks:


GreyCat:
Thanks!

Note, however, that ksy_reference formatting is still severly broken after conversion from our GH wiki. Probably it needs to be slightly restructured, reformatted and partially rewritten ;)

PR #4, opened 2017-04-01
user_guide: add documentation of all methods

davidhicks:
Add documentation to the user guide for methods allowed to be used on
various types. The source of what methods are allowed to be called on
each type is:
  kaitai_struct_compiler/
    shared/
      src/
        main/
	  scala/
	    io/
	      kaitai/
	        struct/
		  translators/
		    BaseTranslator.scala

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -1409,7 +1409,7 @@ No methods are currently supported for floating point numbers.
 
 |`length`
 |Integer
-|Length of a string in number of bytes
+|Length of a string in number of characters
 
 |`reverse`
 |String
@@ -1417,7 +1417,7 @@ No methods are currently supported for floating point numbers.
 
 |`substring(from, to)`
 |String
-|Extracts a portion of a string between byte offset `from` and byte offset `to` (inclusive of bytes at `from` and `to` offsets)
+|Extracts a portion of a string between character at offset `from` and character at offset `to` (inclusive of characters at `from` and `to` offsets)
 
 |`to_i`
 |Integer
@@ -1498,7 +1498,7 @@ a few pre-defined internal methods (they all start with an underscore
 
 |`size`
 |Integer
-|Size of array in number of bytes
+|Number of elements in the array
 |===
 
 ==== Streams
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -1377,30 +1377,82 @@ library" for expression language.
 |Return type
 |Description
 
-|to_s
+|`to_s`
 |String
 |Converts integer into a string using decimal representation
-
-|to_s(radix)
-|String
-|Converts integer into a string using specified `radix` (i.e. use `16` to get hexadecimal representation, use `8` to get octal, etc)
 |===
 
 ==== Floating point numbers
 
-TODO
+No methods are currently supported for floating point numbers.
 
 ==== Byte arrays
 
-TODO
+[cols="3*", options="header"]
+|===
+|Method name
+|Return type
+|Description
+
+|`to_s(encoding)`
+|String
+|Decodes (converts) a byte array encoded using the specified `encoding` scheme into a string
+|===
 
 ==== Strings
 
-TODO
+[cols="3*", options="header"]
+|===
+|Method name
+|Return type
+|Description
+
+|`length`
+|Integer
+|Length of a string in number of bytes
+
+|`reverse`
+|String
+|Reversed version of a string
+
+|`substring(from, to)`
+|String
+|Extracts a portion of a string between byte offset `from` and byte offset `to` (inclusive of bytes at `from` and `to` offsets)
+
+|`to_i`
+|Integer
+|Converts string in decimal representation to an integer
+
+|`to_i(radix)`
+|Integer
+|Converts string with number stored in `radix` representation (i.e. use `16` to get hexadecimal representation, use `8` to get octal, etc) to an integer
+|===
 
 ==== Enums
 
-TODO
+[cols="3*", options="header"]
+|===
+|Method name
+|Return type
+|Description
+
+|`to_i`
+|Integer
+|Converts enum into corresponding integer representation
+|===
+
+==== Booleans
+
+[cols="3*", options="header"]
+|===
+|Method name
+|Return type
+|Description
+
+|`to_i`
+|Integer
+|Returns `0` if the boolean value is `false` or `1` if the boolean value is `true`
+|===
 
 ==== User-defined types
 
@@ -1443,6 +1495,10 @@ a few pre-defined internal methods (they all start with an underscore
 |`last`
 |Array base type
 |Gets last element of the array
+
+|`size`
+|Integer
+|Size of array in number of bytes
 |===
 
 ==== Streams


user_guide.adoc
```diff
@@ -1377,30 +1377,82 @@ library" for expression language.
 |Return type
 |Description
 
-|to_s
+|`to_s`
 |String
 |Converts integer into a string using decimal representation
-
-|to_s(radix)
```
davidhicks, line 1384: I couldn't find evidence in _BaseTranslator.scala_ that _to_s_ can accept arguments for type _IntType_.

Reference: https://github.com/kaitai-io/kaitai_struct_compiler/blob/add79599cc7d0a471b1908f4028ff0080a872f56/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala#L114

GreyCat: This is [exactly it](https://github.com/kaitai-io/kaitai_struct_compiler/blob/add79599cc7d0a471b1908f4028ff0080a872f56/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala#L117) - it accepts one radix argument here.

There's also [an argument-less version](https://github.com/kaitai-io/kaitai_struct_compiler/blob/add79599cc7d0a471b1908f4028ff0080a872f56/shared/src/main/scala/io/kaitai/struct/translators/BaseTranslator.scala#L85) in slightly different place.

davidhicks: That's right, but it only seems to reach that to_s (with radix argument) point in the code if the type which the method is called on is BytesType.

To work with integers, wouldn't we need to add the following:
  case (_: IntType, "to_s") => bytesToStr(translate(obj), args(0))

And then update the various translators to handle conversion of integers into strings with various radixes. For example, CppTranslator.scala currently only implements integer->string conversion if the radix is 10.

davidhicks: To clarify the confusion, this comment thread refers to methods which can be called on IntType, not StrType.

GreyCat: Um, you're right, I'm sorry! I guess we need to have a good solid checklist of methods that need to be implemented, some test suite to check conformance, and, ideally, make it somehow in sync with the documentation automatically...

I wonder if it's a good idea to make it auto-generated somehow...

user_guide.adoc
```diff
@@ -1377,30 +1377,82 @@ library" for expression language.
...
-|String
-|Converts integer into a string using specified `radix` (i.e. use `16` to get hexadecimal representation, use `8` to get octal, etc)
 |===
 
 ==== Floating point numbers
 
-TODO
+No methods are currently supported for floating point numbers.
```
GreyCat, line 1387: Actually, it's again a shame on the compiler. There should be at least `to_i`, `to_s`, that kind of stuff... We totally need a plan...

user_guide.adoc
```diff
@@ -1377,30 +1377,82 @@ library" for expression language.
...
-TODO
+[cols="3*", options="header"]
+|===
+|Method name
+|Return type
+|Description
+
+|`length`
+|Integer
+|Length of a string in number of bytes
```
GreyCat: "of characters", if I see it correctly.

user_guide.adoc
```diff
@@ -1377,30 +1377,82 @@ library" for expression language.
...
+
+|`reverse`
+|String
+|Reversed version of a string
+
+|`substring(from, to)`
+|String
+|Extracts a portion of a string between byte offset `from` and byte offset `to` (inclusive of bytes at `from` and `to` offsets)
```
GreyCat: Again, it is supposed to be operating on character positions, not bytes.

user_guide.adoc
```diff
@@ -1443,6 +1495,10 @@ a few pre-defined internal methods (they all start with an underscore
 |`last`
 |Array base type
 |Gets last element of the array
+
+|`size`
+|Integer
+|Size of array in number of bytes
```
GreyCat: Number of entries in array, actually.

user_guide.adoc
davidhicks: I had only looked at CppTranslator which uses std::string where the length method returns the number of bytes in the string, not the number of UTF-8 codes (characters), and not the number of printable glyphs.

Ignoring the various programming language variants of "string", what is the KS-internal StrType meant to represent? A collection of bytes with unspecified encoding (or perhaps specified elsewhere), 7-bit ASCII encoding, UTF-8 encoding, ...?

user_guide.adoc
davidhicks: Will fix, thanks.

user_guide.adoc
davidhicks: > For std::string, the elements are bytes (objects of type char), which are not the same as characters if a multibyte encoding such as UTF-8 is used. 

From http://en.cppreference.com/w/cpp/string/basic_string/size

davidhicks:
```
#include <string>
 
int main()
{
    std::string s(u8"Exampleب");
    printf("%lu\n", s.size());
    printf("%lu", s.length());
}
```

Returns:
```
9
9
```

Due to the two-byte `ب` Unicode character.

Compiler is GCC 6.1 set to compile as C++17 with post-C++17 changes

GreyCat: "StrType" in general is supposed to represent strings of Unicode characters, to the extent supported natively by the language. That's `String` in Java, `string` in C#, `unicode` in Python, `String` with `encoding` set to `:UTF-8` in Ruby, etc. The whole point of such string types is that you can use it naturally, accessing characters, not bytes, without worrying about any encodings stuff, which is supposed to be accomplished on byte array -> string conversion step.

Unfortunately, situation with C++ is [generally so-so](http://doc.kaitai.io/lang_cpp_stl.html#_string_encoding). There's no universally accepted way to deal with Unicode characters — some people like to use various implementations of wide characters (`wchar_t`, `LPWSTR`, etc), some just store everything in UTF-8 and try to use UTF-8-aware operations, etc, etc.

Given that it wasn't a pressing issue for majority of formats, we've decided to just keep it simple for now and use basic C++ STL methods. Probably the way to go is to have several implementations, selectable using some sort of `#define` in compile-time (i.e. iconv-based implementation with "convert everything to UTF-8 internally" principle, Windows-based implementation using WinAPI stuff, ICU-based, etc, etc).

Given that you use C++, your input here is more than welcome — I'm totally not a C++ expert here :)

davidhicks: Thanks for the explanation, I think I understand the intention of StrType better now.

I'll need to play around with some other file types which do care about string encodings before being able to suggest any improvements. Use of IntType with methods is also something I want to have a look at, as the documentation stating "will return an Integer" doesn't indicate whether it's signed/unsigned, 32bit/64bit, etc.

davidhicks:
Changes made as suggested for array size() and string substring() and string length().

user_guide.adoc
GreyCat: Generally, *a lot* of languages do not care and have a single "integer" type, which is signed and either platform-dependent or infinite (transparently converting to big integer or something like that as it grows). To accomodate that, we just take it as common denominator and state that all returned integers are `CalcIntType`, and they are mapped to some common one-size-fits-most integer type in a target language. And for many languages, it's the only type ;)

Actually, even in languages that do care about many integer types, there's still some degree of int coalescing. For example, if you add `byte` and `byte` in Java, you get `int` automatically. I'm not sure about C++, but I totally won't be surprised if you end up with `int` when trying to add `uint8_t` and `int16_t`, for example.

GreyCat:
Thanks for your contribution!

user_guide.adoc
```diff
@@ -1377,30 +1377,82 @@ library" for expression language.
...
+|Integer
+|Length of a string in number of characters
+
+|`reverse`
+|String
+|Reversed version of a string
+
+|`substring(from, to)`
+|String
+|Extracts a portion of a string between character at offset `from` and character at offset `to` (inclusive of characters at `from` and `to` offsets)
```
GreyCat, line 1420: I'm kind of hesitating about this one... I don't remember what it maps to exactly, need to check if it is really inclusive or not.

GreyCat: `foo.substring(1, 3)` in KSY translates to:

* `foo().substr(1, (3) - (1));` in C++
* `Foo.Substring(1, 3 - 1)` in C#
* `foo().substring(1, 3)` in Java
* `this.foo.substring(1, 3)` in JS
* `$self->foo()[1:3]` in Perl
* `foo[1, (3 - 1)]` in Ruby

I believe it all yields a string starting from character with index=1 (first character has index=0), spanning two characters, i.e. "from" index is included, and "to" is excluded.

PR #5, opened 2017-04-11
User guide documentation update

davidhicks:


GreyCat:
Oh, bummer ;) I've just realized that I wrote streams/substreams section as well on April, 9th, and just forgot to push it into public...

GreyCat:
I'm so sorry... I'll try to think of a way to merge both these sections into one.

davidhicks:
Two sets of documentation is better than zero! :) I'll also see about updating this branch to ensure it can be merged. Is the content of these proposed documentation updates accurate? Feedback appreciated on any inaccuracies or suggestions for rewording.

user_guide.adoc
```diff
@@ -2079,7 +2079,38 @@ beginner Kaitai Struct users.
...
+Where an exception occurs is upon attempting to copy data from the
+`body` substream into the `overflow` object. After data has been
+copied from the `body` substream into the `payload` object, the
+`body` substream will only have 1 byte of data still available for
+the parser to request. As 2 bytes of data are attempted to be
+requested, the `body` substream is exhausted of available data and
+thus an exception occurs. The fact that the root stream still has
+1001 bytes available to be requested from the input file does not
+matter, as the `body` substream never has the opportunity to request
+any more than the first 1000 bytes of the input file.
```
GreyCat, line 2121: This is actually not a pitfall, but a legitimate behavior, and well-explained in previous section.

The "pitfall" I was thinking about in this section is the following: when a new substream is created, all parse instances with positions act within that substream by default.

So, this one works as expected:

```yaml
seq:
  - id: skipped
    size: 1000
  - id: indexing
    type: file_index_entry
    # but adding "size: 24" here will ruin "file_body" instance,
    # although it looks legitimate at the first glance
types:
  file_index_entry:
    seq:
      - id: file_name
        type: str
        size: 16
      - id: file_pos
        type: u4
      - id: file_len
        type: u4
    instances:
      file_body:
        pos: file_pos
        size: file_len
```

To overcome that, one needs to use something like `io: _root._io` in `file_body`. Of course, documentation warrants a somewhat better example and explanation.

user_guide.adoc
```diff
@@ -380,6 +380,21 @@ enums:
     17: udp
 ----
 
+Alternatively, hexadecimal notation can also be used to define an enumeration:
```
GreyCat, line 383: Totally ok, but I'd also noted that this is a service provided by YAML, not something specific to KS.

user_guide.adoc
```diff
@@ -832,6 +832,39 @@ other value which was not listed explicitly.
         _: rec_type_unknown
 ----
 
+If an enumeration has already been defined, you can use references to
+items in the enumeration instead of specifying integers a second time:
```
GreyCat, line 836: Actually, if you defined `key` as enum, then you don't have much choice. You can't compare enums to integers without additional conversions.

user_guide.adoc
```diff
@@ -832,6 +832,39 @@ other value which was not listed explicitly.
...
+types:
+  data_field_width:
+    seq:
+      #...
+  data_field_height:
+    seq:
+     #...
+  data_field_depth:
+    seq:
+     #...
```
GreyCat, line 860: Pedantic person in me cries for that misaligned `#...` ;)
And, anyway, `seq` is totally optional, so may be it's better to wrap it up as:

```yaml
types:
  data_field_width: # ...
  data_field_height: # ...
  data_field_depth: # ...
```

for brevity.

user_guide.adoc
```diff
@@ -1090,7 +1090,117 @@ bytes sparsely.
...
-TODO
+====Introduction and simple example====
+
+A stream is a flow of data from an input file into a parser which is
+generated by a KS script. The parser can request one or more bits of
+data from the stream at a time, but cannot request the same data twice
+and cannot request data be provided out of sequential order. A stream
+knows the maximum amount of data available to be requested by the
+parser and the actual amount of data which has already been
+requested by the parser.
```
GreyCat, line 1101: This explanation is pretty abstract and somewhat misleading. "Stream" can be re-read as many times as needed, and it can be seeked: that's exactly how positional parse instances work, they use `seek` operations on a stream.

user_guide.adoc
```diff
@@ -1090,7 +1090,117 @@ bytes sparsely.
...
+
+When a file is first opened for parsing by a parser generated by KS,
+a root stream is created. This root stream can be accessed via
+`_root._io` at any time and in any place. In this scenario, `_root`
+returns the top level object defined in a script, and `_io` is a
+method which can be called on an object to return the associated
+stream. The root stream will know the maximum amount of data available
+to be requested by the parser as the file size of the input file which
+is being parsed. Initially, the root stream will know that 0 bits of
+data have been requested by the parser.
```
GreyCat, line 1111: Streams can be used on in-memory byte arrays too, not necessarily files (which have file sizes). And, actually, stream does not "know" full file size, but it can query it on demand. File size can change if file is modified when KS parsing is in progress, so it's actually ok to have `_io.size` to return varying values in different points in time.

user_guide.adoc
davidhicks: Excellent. I didn't know about `io:` either, so that's a good one to document! Nice feature!

user_guide.adoc
davidhicks: I was thinking that a new section of the document could be created for general syntax and a very brief overview of YAML and what it provides. This example I provided may be better suited there.

user_guide.adoc
davidhicks: Thanks, agreed

user_guide.adoc
davidhicks: Hmm good point, I'll update the text accordingly

user_guide.adoc
davidhicks: I'll think of another way to explain streams then, especially with reference to how `pos:` works (seeking) and how `io:` can be used to designate which stream to use.

user_guide.adoc
davidhicks: That's a great point, probably one worth adding to the pitfalls section (or troubleshooting or similar) for the few people who may encounter the issue and not understand what is going on.

arekbulski:
This thread is stale, but could you resolve it before I work on any documentation?

GreyCat:
Yeah, I just need to find some time to finish it...

user_guide.adoc
arekbulski: Some Construct features are Python features, but I would advertise them just the same. Purpose of documentation is to show capabilities, not attribution. =) Just saying.

arekbulski:
I will try to help you finish it, with the little understanding of how Kaitai works that I have.

GreyCat:
That would be most awesome %)

arekbulski:
I will carefully review this PR and attempt to solve any existing issues.

PR #6, opened 2017-07-15
Fix typos and make it clear that `contents` only accepts bytes

ghost:
Replacing the word integers with bytes makes it clearer that only positive values between 0 and 255 can be used for the `contents` attribute in my opinion.
This PR also fixes a few typos I noticed and a broken link on the _"Adding support for new target language"_ page.


GreyCat:
Thanks!

PR #8, opened 2017-11-12
Typos correction

AndrewCh:
Typos correction throughout the guide.

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -355,7 +355,7 @@ seq:
 
 === Enums (named integer constants)
 
-The nature of binary format encoding dictates use that in many cases
+The nature of binary format encoding dictates us that in many cases
 we'll be using some kind of integer constants to encode certain
 entities. For example, a widely known IP packet uses 1-byte integer to
 encode protocol type for the payload: 6 would mean "TCP" (which gives us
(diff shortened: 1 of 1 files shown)


user_guide.adoc
```diff
@@ -1434,7 +1434,7 @@ complicated these `pos` expression would get without substreams.
 === Absolute positioning
 
 If you'd ever need to "escape" the limitations of a substream when
-using a `pos` key of a parse instance and address something absolutely
+using a `pos` key of a parsed instance and address something absolutely
```
GreyCat, line 1437: Technically, there are 2 kinds of instances: "parse instances" (which get its results from parsing a stream) and "value instances" (which get their result virtually, by evaluating an expression). Only "parse instances" have `pos` key (as they are indeed positioned somewhere in the stream). It's not a very pretty name, but technically it's there...

GreyCat:
Thanks a lot! I've never ever realized that it's so pestered with typos %) I'm a really lousy writer ;)

I'm not 100% sure about one commented place, but everything else is definitely a good thing :)

user_guide.adoc
AndrewCh: okay. Then you are right.

PR #9, opened 2018-02-04
Reference docs, small updates

arekbulski:
Notice that some of the docs already dropped the "a" "an" in formal descriptions, so this was already a sorta approven writing style. I just aplied it to the remaining text. Also improved one thing or two.


GreyCat:
Thanks!

arekbulski:
Would you approve if I applied same to user guide?
For example in http://doc.kaitai.io/user_guide.html#_checking_for_magic_signatures

GreyCat:
You mean, removal of "a" / "an"? I won't mind, although I don't really see much point in spending lots of time on that relatively minor stuff either %) But go ahead, if you'd like to.

PR #12, opened 2018-03-04
minor edits to Reference docs

arekbulski:
On side note, why are the programmers so nuts about calling everything "foo" and "bar"? Dont they have more imaginative names?

### DIFF
--- a/ksy_reference.adoc
+++ b/ksy_reference.adoc
@@ -746,15 +746,14 @@ by applying type or file-wide default encoding in
 [NOTE]
 ====
 `type: strz` can be also used as a shortcut to define a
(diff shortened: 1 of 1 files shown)


KOLANICH:
```text1``` and ```text2``` are IMHO more boring than `foo` and `bar`. If you wanna more imaginative names, I guess it's better to pick some live examples from the formats library and simplify them. And I guess it's not a good idea to bloat the main repo with PR branches. If you wanna use a PR, put this temporary branch into your own repo rather than into the main one.

GreyCat:
This is purely matter of taste, and I don't think any of the alternatives has strong arguments for it, so we'll leave it as is for now.

PR #14, opened 2018-04-01
non-seekable non-tellable streams explained

arekbulski:
Mentioned in https://github.com/kaitai-io/kaitai_struct_csharp_runtime/issues/8 .

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -3364,3 +3364,26 @@ seq:
 This is because most processing algorithms require to know size of data
 to process beforehand, and final size of `some_body_type` might be
 determined only in run-time, after parsing took place.
+
+=== Supported stream types
+
+There are different types of streams. There are memory-based streams, file-based streams, and special streams (like stdin/stdout/stderr but also sockets pipes block/character devices etc). Those special streams are often non-seekable and non-tellable.
+
+Only certain types of streams are affected:
+* memory-based streams are OK
+* file-based streams are OK
+* stdin/stdout/stderr are affected
+* sockets, pipes, block/character devices are affected
+
+Only some runtimes are affected:
+* C# is affected, `MemoryStream` and `FileStream` are OK but `NetworkStream` and `PipeStream` and `CryptoStream` are not
+* Python is affected, in-memory `BytesIO` and file-based `open()` are OK but special streams are not
+* C++ is OK
+* Java is OK
+
+Only some KSY tags and expressions are affected:
+* expressions using `_io.pos`, `_io.size` and `_io.eof`
+* `seq` fields using `terminator` with `consume: false`
+* `instances` fields using `pos`
+
+Those listed above may still not fail, if they happen to use a substream, which is a memory-based stream which of course, supports seek and tell.


user_guide.adoc
```diff
@@ -3364,3 +3364,16 @@ seq:
...
+
+=== Supported stream types
+
+There are different types of streams. There are memory-based streams, file-based streams, and special streams (like sockets pipes etc). Those special streams are often non-seekable and non-tellable.
+
+Some runtime methods use seek() and/or tell() on the stream. If you are parsing memory-streams or file-streams then those operations are fully supported and there is no problem whatsoever. However, if you are using those special types of streams then be aware that some tags are compiled into a code that uses seek and/or tell, and therefore will fail if the stream does not support those. Most tags do not use seek or tell, and do not care what kind of stream it is.
+
+Methods that involve seek and/or tell:
+* expressions using `_io.pos`, `_io.size` and `_io.eof`
+* seq fields using `terminator` with `consume=False`
```
KOLANICH: Put all the identifiers like `seq`  and `instances` in single back quotes ` to show that they are identifiers, not just words.

KOLANICH (approved):

user_guide.adoc
arekbulski: Fixed.

GreyCat:
The main problem here is that this piece of documentation misses the perspective with all the different languages we have. These claims are mostly true for Python, but, for example:

* in C++, we use [istream](http://www.cplusplus.com/reference/istream/istream/), which is guaranteed to be implementing seekg/tellg, so this problem kind of not exists. If you'll be processing any "special" stream, like packets read from network, you're bound to use that with in-memory buffering.
* in Java, we use two very different approaches — i.e. [RandomAccessFileKaitaiStream](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/master/src/main/java/io/kaitai/struct/RandomAccessFileKaitaiStream.java) and [ByteBufferKaitaiStream](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/master/src/main/java/io/kaitai/struct/ByteBufferKaitaiStream.java), and both are guaranteed to be "seekable" and "tellable". Java indeed has special input streams which are non-seekable (or at least seekable only through buffering & rewinding), but we don't support them at the moment.

etc, etc. So it's all very different and probably warrants rather a language-specific notes, than a general doc section...

arekbulski:
Good point, and I learned something as well. I looked into 2 runtimes (Python and C#) and both used seek/tell in same (corresponding) methods so I assumed that is the same with more or less all runtimes. After all, this section makes it clear that its a generalisation, as in "things may work or not" and "milage may vary". Would it be okay if I added a note about which languages are affected in this (general) guide, instead of separate notes in every per-language guide? It would be easier to write/maintain.

arekbulski:
Also, I think you meant it in the context of C++, but in C# there is a class called [NetworkStream](https://msdn.microsoft.com/en-us/library/system.net.sockets.networkstream(v=vs.110).aspx) which is used by TCP sockets, I think. Honestly I didnt write anything C# since like highschool, some 10 years back. This stream is non-seekable and non-tellable.

arekbulski:
Fixed all outstanding issues, RTM.

GreyCat:
Reviewing it again, I can barely find anything I agree with in this text.

> === Supported stream types

The title is confusing and useless for search purposes. What is "stream type" anyway, and why it suddenly sprung up here? If we want people to quickly find an answer for their problem, we should think of what they would search for in the first place.

> There are different types of streams. There are memory-based streams, file-based streams, and special streams

No, this classification is confusing and just plain wrong. In Java, for example, there are like 5-6 different stream APIs, and we only support some of them. On the other hand, with memory-mapped files, the difference between "memory-based streams" and "file-based streams" is very vague. Not a single programming language we support has something similar to this classification.

> Those special streams are often non-seekable and non-tellable.

Again, "tellable" is virtually unknown for everyone without C knowledge.

> +Only certain types of streams are affected:
> +* memory-based streams are OK
> +* file-based streams are OK
> +* stdin/stdout/stderr are affected
> +* sockets, pipes, block/character devices are affected

Again, this is generalization and oversimplification that does not cover a lot of cases, and, what's more important, it's just plain wrong. Not all "file-based" streams are seekable. Not APIs that allow reading from "sockets, pipes" as a stream and non-seekable. "Block devices" are normally seekable, but that's not guaranteed by API. "Character devices" are usually not seekable, but, again, nothing prevents a certain driver to provide a `lseek` call implementation. For example, debugfs does that. And, again, that is very much OS-specific.

You say that something is "affected", but the text does not even mention what is the problem that "affects" things, and what does that mean in general.

> +* C++ is OK
> +* Java is OK

They are not "OK" or "not OK". That's just how API is designed/implemented. C++ can be "affected" just as well, as you can technically set `std::cin` mode to binary and pass it to `kaitaistream_t`, which would fail with exception on the very first attempt to `seekg()` it. In Java, you can't do that just because we don't provide API to work with generic [InputStream](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html), and `System.in` is just that.

I still think that this whole talk makes sense only in context of certain language and certain stream API. Probably we should add relevant section(s) for per-language notes, which would address:

* How one can create KaitaiStream-like object
  * Using helper functions
  * Manually
* What would be implications of creating that stream from different types of source streams/objects
  * Compatibility with seeking invocations (i.e. restricted at compile-time, fails at run-time, etc)
  * Performance implications
  * Resource consumption implications
  * How that all maps to OS primitives (i.e. opened files, mmap regions, /dev/shm, etc), if that's applicable and possible to predict

Analyzing all that stuff for all languages is a huge piece of work. Let's start with Python, for example, given that it's most familiar to you, and try to replicate the pattern for all other languages eventually?

GreyCat:
Changes request, no movement for over a year. In my opinion, this section warrants almost full rewrite anyway, so closing this for now.

PR #15, opened 2018-04-01
FAQ mentions Construct library

arekbulski:
Mentioned in https://github.com/kaitai-io/kaitai_struct/issues/344 .

### DIFF
--- a/faq.adoc
+++ b/faq.adoc
@@ -28,6 +28,19 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
 
 == How does it compare to ...
 
+=== ... Python library Construct?
+
+Actually, Construct is the closest analog to Kaitai Struct. It is also a declarative and symmetric binary parsing library, but there are significant differences:
+
+* Construct does both parsing and serialization, instead of only parsing (https://github.com/kaitai-io/kaitai_struct/issues/27[feature #27]).
+* Construct is a Python-only module, instead of supporting multiple languagues.
+* Construct is "declarative" in sense that it defines data structures instead of parsing code. The structures are still defined using Python language, instead of YAML.
+* Construct aims at offering more sophisticated building blocks, including those only available on Python like Pickle and Numpy protocols, instead of most basic/common elements.
+
+In fact, there is an open and active collaboration between Construct and Kaitai Struct. There are (currently being implemented) import/export tools, that allow translating schemes between the two frameworks.
+
+Main documentation site: https://construct.readthedocs.io
+
 === ... Google Protocol Buffers, ASN.1, Apache Thrift, Apache Avro, BSON, etc?
 
 They're completely different. Projects mentioned are actually different serialization specifications that map existing data into some sort of extensible binary stream, usually for easy transmission / interchange. Binary representation is driven by the data and encoded according to particular standard of a given protocol, which usually has a fixed representation for integers, for strings, for arrays, for dictionaries, etc. Most of these project allow generated formats to be automatically extensibile, carry versioning information, automatically embed typing information of some sort.


faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
 
 == How does it compare to ...
 
+=== ... Python library Construct?
+
+Actually, Construct is the closest analog to Kaitai Struct. It is also a symmetric binary parsing library, but the difference is mostly that:
+
+* Construct does building and parsing while Kaitai (currently) only does parsing
```
KOLANICH: `building` is called [`serialization`](https://github.com/kaitai-io/kaitai_struct/issues/27) in KS.
>currently

should be a link to the issue

arekbulski: Fixed.

faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
 
 == How does it compare to ...
 
+=== ... Python library Construct?
+
+Actually, Construct is the closest analog to Kaitai Struct. It is also a symmetric binary parsing library, but the difference is mostly that:
+
+* Construct does building (serialization) and parsing while Kaitai ([currently](https://github.com/kaitai-io/kaitai_struct/issues/27)) only does parsing
```
KOLANICH: maybe
>does both

KOLANICH: And I guess that rewriting commits to keep the commit tree clean is fine if these commits are not cloned by anyone.

arekbulski: Fixed and amended.

faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
 
 == How does it compare to ...
 
+=== ... Python library Construct?
+
+Actually, Construct is the closest analog to Kaitai Struct. It is also a symmetric binary parsing library, but the difference is mostly that:
```
GreyCat: Line lengths are not the same as in the rest of the file.

faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
...
+
+* Construct does both parsing and serialization while Kaitai ([currently](https://github.com/kaitai-io/kaitai_struct/issues/27)) does only parsing
```
GreyCat: We never call KS "Kaitai", especially in official documentation. Also, this is not Markdown, it's AsciiDoc, so this link won't work. It should be formatted like

```
https://github.com/kaitai-io/kaitai_struct/issues/27[currently]
```

Which, in turn, shows that you haven't even bothered to check before submitting.

faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
...
+* Construct is a Python only module while Kaitai supports a multitude of target languages
```
GreyCat: It's spelled "Python-only" in English.

faq.adoc
```diff
@@ -28,6 +28,15 @@ So far Kaitai Struct focuses on reading (parsing) only. There are plans to suppo
...
+
+Main documentation site: https://construct.readthedocs.io
```
GreyCat: The rest of these sections use a breakout boxes to present this link, why don't we follow the suit here?

GreyCat:
I have lots of questions for just about every line here. On the other hand, this section misses lots of stuff which is probably important, such as:

* The emphasis on the fact that both languages are declarative, but with a short explanation of different understanding of "declarativity" in KS and Construct
* A mention of that Construct aims at a higher-level conversion primitives, such as CString, TimeStamp, etc.
* List of Construct features which are missing in KS
* A mention of KS <-> Construct collaboration
  * A mention of Construct -> KS conversion tool
  * A mention of KS -> Construct target

We're indeed doing lots of interesting and unique things here, let's show them off to people :)

GreyCat (changes requested):

faq.adoc
arekbulski: You are kinda wrong on this one. I didnt compile the docs because I dont know how. I did wonder how to express links, so looked in the remaining content and found this syntax so I used it. Sorry but I never worked with this format before.

faq.adoc
arekbulski: Right :)

faq.adoc
arekbulski: Uhm would you want me to add a box? Okies.

arekbulski:
> I have lots of questions for just about every line here. On the other hand, this section misses lots of stuff which is probably important, such as:

Good points! Those are worthy points to make, it just didnt occur to me at the time. I will add some content and request re-review.

faq.adoc
GreyCat: Building the docs is essentially a matter of running "make", once you'll get asciidoctor (or actually any other asciidoc processor of your choice) installed: https://github.com/kaitai-io/kaitai_struct_doc/blob/master/.travis.yml#L16-L20

Sorry about the mess with the other links, I'll try to fix them shortly.

arekbulski: Maybe we could add a note in those "how to build compiler/run test" guide?

faq.adoc
arekbulski: The file has infinite lines.

faq.adoc
arekbulski: Fixed.

faq.adoc
arekbulski: Fixed.

faq.adoc
arekbulski: Preon section doesnt use boxes, and I would rather not do that.

arekbulski:
Fixed all issues, amended commit, RTM.

GreyCat:
Ok, that's a start :)

PR #18, opened 2018-11-01
Update maven dependency to correct version.

brunchboy:
I was trying to adopt this tool in a Java project, and some classes in
the code generated by `ksc` were not found. I eventually thought to go
to Maven Central and, sure enough, found that there was a newer
release than the one the documentation told me to use.

You might want to consider using a tool like Maven Badges,
https://github.com/softwaremill/maven-badges so that your
documentation file is rendered with an image that shows the actual
latest release version, and which when clicked takes you to a page
showing how to add it as a dependency using many popular JVM-hosted
build tools. I do this on my own project pages, for example:
https://github.com/Deep-Symmetry/beat-link#installing

### DIFF
--- a/lang_java.adoc
+++ b/lang_java.adoc
@@ -41,7 +41,7 @@ following to one's `pom.xml`:
 <dependency>
     <groupId>io.kaitai</groupId>
     <artifactId>kaitai-struct-runtime</artifactId>
-    <version>0.7</version>
+    <version>0.8</version>
 </dependency>
 ----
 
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks!

I guess using a badge is great idea, please feel free to submit a PR to add it. However, I'd like to keep text documentation up to date too. Currently we have [this workflow](http://doc.kaitai.io/developers.html#_java) documented, probably it warrants adding a documentation update to it as well.

PR #20, opened 2018-11-21
Add suggestion to match top-level ID with filename

brunchboy:
Based on a recommendation from @KOLANICH

Also fixed a typo I noticed while searching to see if this was
already explained anywhere.

### DIFF
--- a/ksy_style_guide.adoc
+++ b/ksy_style_guide.adoc
@@ -106,7 +106,7 @@ different purposes.
 
 For top-level meta section, use the following order of keys:
 
-. `id` — main identifier
+. `id` — main identifier, should match the name of the `.ksy` file
 . Human-readable meta-information (in order of importance, at least
   one MUST be present):
 .. `title`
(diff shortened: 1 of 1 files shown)


ksy_style_guide.adoc
```diff
@@ -106,7 +106,7 @@ different purposes.
 
 For top-level meta section, use the following order of keys:
 
-. `id` — main identifier
+. `id` — main identifier, should match the name of the `.ksy` file
```
GreyCat: Totally makes sense, but please use `MUST` instead of `should`, as this is actually a pretty hard requirement.

GreyCat: (and it is `MUST`, not `must`, as we're reffering to RFC 2119 terms).

brunchboy: Certainly! This change has been added to the pull request.

GreyCat:
Thanks!

PR #21, opened 2018-11-21
Add warning about incomplete FourCC enum crashes

brunchboy:
Here is a possible version of the warning. I added a link to the underlying issue, which will hopefully help people figure out when it is safe to do this even if we don’t immediately remember to update the user guide when it is fixed.


GreyCat:
Thanks!

PR #26, opened 2019-04-11
Fix for the sample code in choosing a substream

tlindner:
Line 25 of the sample code should be "io: _root.block_one._io", not "io: _root.big_container._io".


GreyCat:
Thank you, you're absolutely right! Sorry for this taking so long to commit.

PR #28, opened 2019-06-27
Added line to disambiguate working directory for "Building for JVM".

GabriellaNicoleRamirez:
Hi all,

As the team and I were looking at adding language support, we ran into a problem where we didn't realize we needed `kaitai-struct/compiler` as our working directory, instead of simply `kaitai_struct`. I just added a line in the documentation that I hope clarifies that for future users.

Thanks!


GreyCat:
Thanks!

PR #29, opened 2019-08-15
Parantheses in bit-sized integers are not correct

StefanD986:
I propose to change the documentation so that its clear that the parantheses should be replaced. 
Additionally I added a small example.

### DIFF
--- a/ksy_reference.adoc
+++ b/ksy_reference.adoc
@@ -680,8 +680,8 @@ For the sake of completeness, here's the full table of available integer types:
 ### Bit-size integers
 
 To specify integers having non-standard number of bits in them, one
-can use the following pattern: `b(\d+)`, where `\d+` is the number of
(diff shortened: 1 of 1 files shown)


generalmimon:
I'm closing this PR, because the KSY reference is now deleted _(as it contained a lot outdated info)_: https://github.com/kaitai-io/kaitai_struct_doc/pull/42

Maybe it could be somehow incorporated to the [`ksy_schema.json`](https://github.com/kaitai-io/ksy_schema/blob/cb0b486f0595f827b51d8f559ed8fc5c0374d5f9/ksy_schema.json#L356) file, which now serves as a data source for the KSY diagram at https://doc.kaitai.io/ksy_diagram.html, but that would require some creative approach, something like putting a table like this to the description of `{seq,instances}/X/type` key:

image

PR #32, opened 2020-01-31
Python packaging-related fixes:

### LINKED ISSUE #692: no python pre-built SNAPSHOT available
In order to test if an issue was fixed in master:
- I installed the 0.9-SNAPSHOT from https://dl.bintray.com/kaitai-io/debian_unstable/ as described on http://kaitai.io/#download
- I ran `pip3 install kaitaistruct`, but only stable releases are posted (link: [PyPi kaitaistruct history](https://pypi.org/project/kaitaistruct/#history)) so 0.8 is the latest Python API

Of course, the two are not compatible which causes an exception:
```
Exception: Incompatible Kaitai Struct Python API: 0.9 or later is required, but you have 0.8
```
As a developer, if I choose to use 0.9-SNAPSHOT when testing, instead of building from source, then it would be nice to have corresponding SNAPSHOT for the other items like Python API.

KOLANICH:
* improved the docs on installing python runtime
* fixed the links to kaitaiStructCompile.setuptools
* added more examples of usage of kaitaiStructCompile.setuptools

Closes https://github.com/kaitai-io/kaitai_struct/issues/692.


lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
 (i.e. if you use ksc v0.8, then you must use compatible v0.8 runtime).
 
 [[add-runtime-release]]
-==== Using release version
+==== Using release version (NOT recommended)
```
GreyCat: Can you elaborate, how exactly is using release is "not recommended"?

KOLANICH: I personally don't recommend since they are outdated most of the time.

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
+of your `setup.cfg` (or the arg of `setuptools.setup`
+with the same name in your `setup.py`) :
 +
 ----
 kaitaistruct
 ----
 +
-and then run `pip install -r requirements.txt` to update all your
-dependencies.
+and then run `pip install --upgrade -e ./` in the dir with the `setup.py`
```
dgelessus: Please use `python3 -m pip ...` instead of `pip ...`: https://snarky.ca/why-you-should-use-python-m-pip/

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
 
 If you're using latest (unreleased AKA "unstable" AKA "snapshot") build
 of Kaitai Struct compiler, that will require latest runtime libraries as
 well:
 
-* If your project is using *requirements.txt* to manage your
-  dependencies, add the following line to it:
+* You can a make `pip` to use a bleeding edge version of the runtime by
+  adding a https://www.python.org/dev/peps/pep-0508/[PEP 508]
+  dependency specifier into `install_requires`:
```
dgelessus: I don't think `pip` is relevant here - there are other tools that handle dependencies and install packages too. I would instead write "You can use a bleeding edge version of the runtime ..." without mentioning `pip` here.

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
+.
 
 * Otherwise, you can just install the package manually using *pip*
   (Python package manager):
 +
 [source,shell]
 pip install --upgrade --pre git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
 
+[[secure-installation]]
+==== More secure systemwide installation (RECOMMENDED)
```
dgelessus: There are reasons why you may want to install the library from a manual clone of the repo, but I don't think security has much to do with it. As far as I know, the `git+https://...` dependency specifier does exactly the same thing behind the scenes (clone the repo, build a wheel and install it). The only difference is that with a manual clone you can look at the contents of the repo beforehand, but realistically most people don't check their dependencies that thoroughly.

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
+
+```sh
+# run unelevated:
+git clone https://github.com/kaitai-io/kaitai_struct_python_runtime.git
+cd kaitai_struct_python_runtime
+python3 ./setup.py bdist_wheel
+# inspect the content of the zip archive in ./dist
+
+# run elevated:
+sudo pip3 install --upgrade ./dist/*.whl
```
dgelessus: You really shouldn't `sudo pip install` anything, especially not when you're worried about security :) The rest of the guide assumes that you already have sufficient privileges to use `pip install`, so I don't think it's a good idea to introduce `sudo` here.

Also, you can run `python3 -m pip install --upgrade ./` (without `-e`/`--editable`) to build and install a wheel of the project in a single step. If you leave out `sudo`, it's no longer a security problem to have `pip` run the `setup.py`.

lang_python.adoc
```diff
@@ -103,9 +129,14 @@ available out of the box in standard libraries since Python v3.4, or
 since v2.4 using https://pypi.python.org/pypi/enum34[enum34] PyPi
 compatibility layer.
 
-* On Debian / Ubuntu, this library can be installed with `sudo apt-get
-  install python-enum34`
-* Otherwise, one can use just `sudo pip install enum34`
+* One usually doesn't need to install this lib manually,
+   `pip` will install it automatically if it is needed.
```
dgelessus: Is that actually the case? AFAIK, kaitai_struct_python_runtime doesn't depend on `enum34`, so there's no reason why it would be installed automatically.

lang_python.adoc
```diff
@@ -103,9 +129,14 @@ available out of the box in standard libraries since Python v3.4, or
...
+* One can also install it manually:
+** Otherwise, one can use just `sudo pip install enum34`. Don't do it on later pythons!
```
dgelessus: Again, we shouldn't be recommending `sudo pip install` here.

lang_python.adoc
```diff
@@ -103,9 +129,14 @@ available out of the box in standard libraries since Python v3.4, or
...
+** On Debian / Ubuntu, this library can be installed with `sudo apt-get install python-enum34`
+* If you have to use 3.4 you may want to replace `enum.py` with the one from
+   https://raw.githubusercontent.com/python/cpython/3.6/Lib/enum.py[python 3.6]
+   in order to get `IntFlag` (not currently used in KS, but may be used in future).
```
dgelessus: This is bad advice - you really shouldn't be replacing files in your Python installation's stdlib. If you need a newer `enum` module and for some reason can't upgrade to 3.6, you should copy the new `enum` module to `site-packages` (or some other directory on `sys.path`), not into the stdlib directory.

In any case, considering that KS currently doesn't use any 3.6 `enum` features, I think this advice shouldn't be added to the documentation yet.

lang_python.adoc
```diff
@@ -114,30 +145,35 @@ https://setuptools.readthedocs.io/en/latest/[Setuptools] is the de-facto
...
 
 Please note that this lib is NOT available on
 https://pypi.python.org[pypi] and if you want to install it
 automatically, you should add
-`git+https://github.com/KOLANICH/kaitaiStructCompile.py` into
-`dependency_links` in the config passed to `setuptools.setup`.
+`kaitaiStructCompile.setuptools @ https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools` into `setup_requires`
+into `options` section of your `setup.cfg`
+and into `requires` of `build_system` section
+of your `pyproject.toml.
```
dgelessus: The \` is missing after `pyproject.toml`.

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
+  dependency specifier into `install_requires`:
 +
 ----
-kaitaistruct @ git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
+    kaitaistruct @ git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
 ----
 +
-and then run `pip install -r requirements.txt` to update all your
-dependencies.
+.
```
dgelessus: I think this line can be removed? Or does this have some special meaning in asciidoc that I'm not aware of?

lang_python.adoc
GreyCat: Backtick missing?

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
 +
 [source,shell]
-sudo apt-get install python-kaitaistruct
+sudo apt-get install python3-kaitaistruct
 +
-Reference: https://packages.debian.org/buster/python-kaitaistruct[package info for Debian], https://packages.ubuntu.com/eoan/python-kaitaistruct[package info for Ubuntu]
+Reference: https://packages.debian.org/buster/python3-kaitaistruct[package info for Debian], https://packages.ubuntu.com/eoan/python3-kaitaistruct[package info for Ubuntu]
 
-** For Python 3.x:
+** For Python 2.x (DEPRECATED, python 2 is EOL since 2020-01-01):
```
GreyCat: Let's actually move this Python version debate into separate section above everything? That section could contain a list of officially supported versions + a note about depreciation?

lang_python.adoc
KOLANICH: >Please use `python3 -m pip ...` instead of `pip ...`: https://snarky.ca/why-you-should-use-python-m-pip/

IMHO, one should just do

```bash
git clone --depth=1 https://github.com/pypa/pip.git
cd pip/src
<needed_python_interpreter> ../setup.py bdist_wheel
<needed_python_interpreter> -m pip install --upgrade ./dist/*.whl
```

this will make `pip` command to use the needed interpreter without `<needed_python_interpreter> -m`.

lang_python.adoc
KOLANICH: >As far as I know, the git+https://... dependency specifier does exactly the same thing behind the scenes (clone the repo, build a wheel and install it).

There is a nuance: all the stages have to be done from root this way.

lang_python.adoc
KOLANICH: >You really shouldn't sudo pip install anything, especially not when you're worried about security :)

Of course I should if I do it with care. Threat model: there is no malware on my PC. Mitigations: disable pip as a source of packages, examine packages source code.

>Also, you can run python3 -m pip install --upgrade ./ (without -e/--editable)

A user usually needs an editable installation when working on a project.

lang_python.adoc
KOLANICH: Damn, I have forgotten to make a PR about it.

lang_python.adoc
KOLANICH: >This is bad advice - you really shouldn't be replacing files in your Python installation's stdlib.

Why not? it works.

>In any case, considering that KS currently doesn't use any 3.6 enum features, I think this advice shouldn't be added to the documentation yet.

OK

lang_python.adoc
KOLANICH: Gonna fix, thanks.

lang_python.adoc
```diff
@@ -33,68 +33,94 @@ Runtime library API version must match compiler version that you're using
...
+sudo pip3 install --upgrade ./dist/*.whl
+```
+
+This is secure only because wheels cannot
+execute commands during installation.
+Though you still need to check what you install.
+
 [[add-dependencies]]
-=== Adding dependencies
+=== Dealing with old python versions
```
GreyCat: Please don't rename these sections, especially without changing the anchor. The idea is to have uniform set of sections, the same ones for all languages.

dgelessus: > The idea is to have uniform set of sections, the same ones for all languages.

That part isn't obvious IMO - maybe it should be documented somewhere? Not sure what would be the best place for that kind of information though - maybe a short comment at the top of every `lang_*.adoc` file?

lang_python.adoc
KOLANICH: It's just here to add a full stop.

lang_python.adoc
KOLANICH: `pip` is the de-facto standard. Other package managers like `conda`, `apt`, `dnf` are not native to python ecosystem and don't use PEP 508.

lang_python.adoc
KOLANICH: Should I change the anchor or should I revert the rename?

lang_python.adoc
dgelessus: I'm not sure how these steps are supposed to help here? The main reason why you're supposed to use `python -m pip` instead of `pip` is that there's no guarantee that `pip` refers to the same Python interpreter as `python`. There are other reasons too, which are explained in the blog post that I've linked. The steps that you've suggested don't help against any of those problems.

Besides, installing things directly from Git for no reason is a terrible idea. You have no idea what state the repo is currently in - it's most likely a development version that could have all sorts of bugs, incomplete features or breaking changes. You also have to run setup.py, which you almost never have to do when installing from PyPI, because every decent package provides pre-built wheels. Also, any depenencies of the project will by default be installed from PyPI anyway, which makes the whole process kind of pointless.

lang_python.adoc
dgelessus: There are lots of other build tools that are native to the Python ecosystem, and some of them (e. g. [flit](https://github.com/takluyver/flit)) even use PEP 508 dependency specifiers.

It's not just about `pip` though - IMHO the entire sentence goes into too much detail that isn't really relevant here. I think it would be enough to say "You can use a bleeding edge runtime by adding this dependency to your `install_requires`: ..." without talking about pip and PEP 508.

lang_python.adoc
dgelessus: The instructions before this one don't use `sudo` (and they shouldn't, see my other comments below).

lang_python.adoc
dgelessus: > Of course I should if I do it with care.

No, really, you should never use `pip` with `sudo`. If you think you need `pip` with `sudo`, you probably want `pip` with `--user`, or a virtual environment. Using `pip` with `sudo` tends to mess with your distro's package manager and possibly break system tools that depend on Python packages.

> disable pip as a source of packages

That's not what you're doing here though. Installing a wheel will still pull in dependencies from PyPI, potentially triggering a setup.py run. Example:

```sh
$ python3 -m pip install sympy-1.5.1-py2.py3-none-any.whl
Processing ./sympy-1.5.1-py2.py3-none-any.whl
Collecting mpmath>=0.19
  Using cached mpmath-1.1.0.tar.gz (512 kB)
Building wheels for collected packages: mpmath
  Building wheel for mpmath (setup.py) ... done
  Created wheel for mpmath: filename=mpmath-1.1.0-py3-none-any.whl size=532239 sha256=911caa345e04f60276e2362c280add58d21437dda5043f0bc27150b19a9ec26c
  Stored in directory: .../Caches/pip/wheels/29/2c/1c/d2e4580cde2743b0aef389e936ac21a2db92921ddbca53faa1
Successfully built mpmath
Installing collected packages: mpmath, sympy
Successfully installed mpmath-1.1.0 sympy-1.5.1
```

I didn't use `sudo` here obviously, but if I had, the setup.py of `mpmath` would have run as root.

> A user usually needs an editable installation when working on a project.

That's not the point though. Your current instructions (that I've commented on) don't do an editable install, so my suggested replacement command doesn't either. These instructions are for users of KS, not for developers, so I don't think editable installs are necessary here.

lang_python.adoc
dgelessus: > Why not? it works.

Because you're replacing the existing standard `enum.py`, so there's no good way to revert the change without reinstalling Python. Putting the file into `site-packages` is also not *great* (because then you have a file in `site-packages` without any package metadata, so pip won't know what to do with it or how to uninstall it), but it's still better than replacing the original `enum.py`.

lang_python.adoc
KOLANICH: >The main reason why you're supposed to use python -m pip instead of pip is that there's no guarantee that pip refers to the same Python interpreter as python.

When you install pip yourself you control which interpreter will be used - the one you use to install `pip` itself.

>Besides, installing things directly from Git for no reason is a terrible idea.

IMHO it is a very good idea.

* On git the live code resides, that is supposed to be seen by the ones developing it reguralily. On pypi there are zip archives almost noone looks into. 

* One gets the latest features and bug fixes. I have been used PEP 508 in my packages a half of a year before this feature has landed to a release.

* >You have no idea what state the repo is currently in - it's most likely a development version that could have all sorts of bugs, incomplete features or breaking changes.

Then I want to encounter them before they are standardized as immutable features that will be preserved till the end of the Universe. It is always YOUR fault, if YOUR package is incompatible with the latest version of its dependency. One day that version will land and your users will have to downgrade it because of your package. The general rule: you can always expect people to upgrade the software when the upgrade is free (not requiring buying anything or rewriting anything from scratch or doing some costly research or doing some costly computations, I hope you understand me), but you cannot expect people to downgrade software because they may rely on new features. `venv`s are not a solution, they are just shit some lazy people use hoping that they will help them avoid upgrading their and third-party unmaintained code.

So some simple rules to follow:
1. always use the latest versions where it is feasible
2. always make your code compatible to the latest versions
3. if you encounter regressions because of bugs in the dependencies, contact their maintainers immediately and insist on fixing
4. if the maintainers are uncooperative, you have to find new dependencies (for example an own fork)

>You also have to run setup.py, which you almost never have to do when installing from PyPI, because every decent package provides pre-built wheels.

For pip it is feasible (it just zips the contents of the repo + adds some metadata). For numpy it is semi-feasible (requires some time for compilation). For anything involving source-customized LLVM it is closer to infeasible (several hours of compilation without any other activity). For tensorflow or plaidml or pytorch or Android or a new build of Firefox or LibreOffice or any other software that takes a lot of RAM to build or a lot of time to build or something using Google Bazel shit that rebuilds everything it is completely infeasible.

>Also, any depenencies of the project will by default be installed from PyPI anyway, which makes the whole process kind of pointless.

It is disableable.

lang_python.adoc
KOLANICH: >There are lots of other build tools that are native to the Python ecosystem

They are not native, most of packages still use `setuptools`, and I hope they will never be native in the sense 1/4 of packages using `flit`, 1/4 using `poetry` shit, 1/4 using `setuptools` and 1/4 using something else. I already have issues when some maintainers have chosen to use `poetry`, which has an incompatible metadata format and has a principial position to use `venv`s I don't want to touch anyhow because they are damn slow, eat hdd space and cause mess in managing the packages installed in them. So I had to rewrite the config into setup.cfg (unfortunately I haven't managed to make `dephell` work). Standardizing `pyproject.toml` without standardizing common metadata and build recipy format was a major mistake that will haunt us for a long time.

The best we can do is to complete ignore the tools made intentionally incompatible with `setuptools` as much as possible.

lang_python.adoc
KOLANICH: If you want to install `pip` systemwide (you usually want, if you want to use it to install systemwide (from root), and you want to install packages systemwide if you want to run some of them as root, and you want to run some of them as root if they are intended to be run this way, for example a package manager written in python or a package messing with other files or devices that are properly accessible to processes run as root only), you have to use `sudo`.

lang_python.adoc
KOLANICH: >Because you're replacing the existing standard enum.py, so there's no good way to revert the change without reinstalling Python.

If one considers he may want to revert, he can keep a backup. I have kept a backup for the case of 3.6 -> 3.4, but everything works fine.

Of course in an ideal case the standard lib of python is just a set of preinstalled packages that can be upgraded independently of python. But it hasn't been done yet, probably they consider it as not worthing the effort :(.

lang_python.adoc
KOLANICH: >Using pip with sudo tends to mess with your distro's package manager and possibly break system tools that depend on Python packages.

Not break, but improve. Trust my experience :) Anyway, if something goes broken, it is a strong reason to fetch its sources and fix them.

lang_python.adoc
dgelessus: > When you install pip yourself you control which interpreter will be used - the one you use to install `pip` itself.

That's not how this works. Installing `pip` (or any other package) does not change your PATH. If the `pip` command refers to a different interpreter than you need/expect, then reinstalling the `pip` package changes nothing about that - the `pip` command will still refer to exactly the same (wrong) interpreter as before.

> IMHO it is a very good idea.

OK, if that workflow works for you, great - I'm not stopping you from installing packages this way on your system. But the usual way to install dependencies in Python is by having your dependency management tool download pre-packaged release versions from PyPI. The KS Python guide should follow that established standard, unless there's a very good reason why KS specifically needs a different workflow (which is not the case as far as I can tell).

> For pip it is feasible (it just zips the contents of the repo + adds some metadata). For numpy it is semi-feasible (requires some time for compilation). For anything involving source-customized LLVM it is closer to infeasible (several hours of compilation without any other activity).

I'm not sure what you mean here... It's up to the package maintainers to build wheels for PyPI, so as a user you don't need to worry about how long the build takes. All of the Python packages you named provide Windows, Mac and [manylinux1](https://www.python.org/dev/peps/pep-0513/) x86_64 wheels (which covers the vast majority of users), so it seems to be feasible in most cases.

lang_python.adoc
dgelessus: > They are not native

By what definition? All of these build tools are compatbile with the pyproject.toml `[build-system]` section as specified in [PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/), so they can be automatically called by other build tools to produce sdists and wheels with standard metadata.

It's not really relevant whether these build tools are compatible with setuptools. The entire point of pyproject.toml is to move away from setup.py (which is based on distutils/setuptools) as the standard build entry point, so that new build can be developed that don't rely on setuptools.

lang_python.adoc
dgelessus: > Not break, but improve. Trust my experience :) Anyway, if something goes broken, it is a strong reason to fetch its sources and fix them.

Again, if that works for you, great. Most people just want to use KS though, and don't want to spend time fixing their distro's Python installation. The majority of users doesn't need to run KS-based code as root, so installing with `--user` is not an issue.

lang_python.adoc
KOLANICH: >By what definition?

By definition that I still can install most of packages without installing a zoo of build systems using only `setuptools`.

lang_python.adoc
```diff
@@ -40,72 +48,80 @@ methods:
...
 ==== Using latest (snapshot) version
 
-If you're using latest (unreleased AKA "unstable" AKA "snapshot") build
+If you're using latest (unreleased AKA "unstable" AKA "snapshot" AKA "nightly") build
 of Kaitai Struct compiler, that will require latest runtime libraries as
 well:
 
-* If your project is using *requirements.txt* to manage your
-  dependencies, add the following line to it:
+* You can a make `pip` to use a bleeding edge version of the runtime by
```
generalmimon: _"Bleeding edge"_ seems a bit tabloid to me, what about "development version" or "the most recent" version?

KOLANICH: Will `version, incorporating the most recent time-untested changes` be any better?

lang_python.adoc
```diff
@@ -114,30 +130,33 @@ https://setuptools.readthedocs.io/en/latest/[Setuptools] is the de-facto
...
 `kaitai` to your `setup.py` where you enumerate files that you want
 compiled, output directory, compilation flags, and source code
 post-compilation transformations. Here are some examples:
-https://github.com/KOLANICH/NTMDTRead/blob/master/setup.py[1],
-https://github.com/KOLANICH/SpecprParser.py/blob/master/setup.py[2].
+https://gitlab.com/KOLANICH/NTMDTRead/blob/master/setup.py[1],
+https://gitlab.com/KOLANICH/SpecprParser.py/blob/master/setup.py[2],
+https://gitlab.com/KOLANICH/FrozenTable.py/blob/master/setup.py[3],
+https://gitlab.com/KOLANICH/MFCTool.py/blob/master/setup.py[4],
+https://gitlab.com/KOLANICH/RDataParser.py/blob/master/setup.py[5].
```
generalmimon: These numbered links look ugly when are in the same line separated by commas (_1, 2, 3, 4, 5_). Another problem with them is that listing "configuration examples" like this _is like a box of chocolates: you never know what you're going to get_. This is quite silly format for the docs of your project: if one would consider using it, is he supposed to open all these examples and manually compare them so he knows how does one use this library in practice?

It would be a lot better if you'd put them into a normal list, one example by line, in format "link - description" (or better "name - description"). Every example should have a unique short description, you don't have to describe the project that the `setup.py` example is from, but rather what features of the `setup.py` provided by your library it uses.

However, I'd rather move these examples into the actual documentation of your [`kaitaiStructCompile.py` repository](https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.py). This could provide you a lot of options how to "sell" these examples - you can paste an example snippet of a possible `setup.py` configuration into the `README.md` of your project, describe in a few sentences what it does, then paste another `setup.py` from another project, point what's different from the previous one, and so on.

This numerical series IMHO isn't much helpful here, and even if it was a "name - description" list, I'm still not sure if it makes sense to have it here in the KS Python notes.

lang_python.adoc
generalmimon: > version, incorporating the most recent time-untested changes

This phrase can be unambiguously replaced with a widely accepted term _unstable_.

KOLANICH: `unstable` has ambigious meaning:

1. changing frequently in incompatible ways.
2. crashing constantly
3. mad, made by mentally unstable people :)

I guess the latter ones don't apply here. And the first one can hardly be applied.

lang_python.adoc
```diff
@@ -17,21 +17,29 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+with both Python 2 and 3. It is strongly recommended to use python 3
+because python 2 is EOL since 2020-01-01.
 
-TODO: document `--python-package`
+`--python-package` allows you to select a parent package name used by
+one generated file to import symbols from another one. If you want to
+embed KSC-generated files into an own package, it is recommended to
+set it to `.`, which means "current package" rather than to the real name
+of the subpackage, because setting it to a real name will cause breakage
+when you want to reorganize or rename your package.
```
generalmimon: First information should be what the `--python-package` is at all and how one can use it, before you tell what it _allows_. It should be mentioned here that it's a command line option that one can supply to the `kaitai-struct-compiler` command, and also some common usage examples would come in handy. It can be also already present as "optional parameter" in the first command like `kaitai-struct-compiler -t python my_spec.ksy [ --python-package whatever ]` so it's clear.

lang_python.adoc
```diff
@@ -17,21 +17,29 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
 Code generated by Kaitai Struct compiler will require additional runtime
-library (https://pypi.org/project/kaitaistruct/[kaitaistruct at PyPi],
+library (https://pypi.org/project/kaitaistruct/[image:https://img.shields.io/pypi/v/kaitaistruct.svg[kaitaistruct at PyPi]],
 https://github.com/kaitai-io/kaitai_struct_python_runtime[sources at
 GitHub]).
 
 Runtime library API version must match compiler version that you're using
 (i.e. if you use ksc v0.8, then you must use compatible v0.8 runtime).
 
+If you are using non-release KSC, you should also use non-release runtime.
```
generalmimon: I don't like the term non-release here, IMHO something like "latest development" would do better.

lang_python.adoc
generalmimon: What about the development version?

KOLANICH: A bit better, but doesn't reflect the fact it can contain undiscovered and undocumented bugs with a bit higher probability.

lang_python.adoc
generalmimon: It would be also cool to show what it actually translates to in the generated Python code. For example, you can show that passing `--python-package ...` causes insertion of line `package ...` (don't actually know, just guessing).

People familiar with Python will rather understand what it's supposed to do when they get some connection to their language.

And if you can rather add more Python snippets instead of this long paragraph, even better :+1:

lang_python.adoc
```diff
@@ -40,72 +48,80 @@ methods:
...
 ----
 kaitaistruct
 ----
 +
-and then run `pip install -r requirements.txt` to update all your
-dependencies.
+and then run `python3 -m pip install --upgrade -e ./` in the dir with your project
+in order to install your package and all its dependencies in
+an editable way. If you don't want to package your software,
+but only installbits dependencies, add `kaitaistruct` into `requirements.txt`.
```
generalmimon: Probably a typo:
```suggestion
but only install its dependencies, add `kaitaistruct` into `requirements.txt`.
```

lang_python.adoc
generalmimon: OK, feel free to keep it if you want, it's not so important.

lang_python.adoc
```diff
@@ -40,72 +48,80 @@ methods:
...
 
-* If your project is using *requirements.txt* to manage your
-  dependencies, add the following line to it:
+* You can a make `pip` to use a bleeding edge version of the runtime by
+  adding a https://www.python.org/dev/peps/pep-0508/[PEP 508]
+  dependency specifier into the config of your favorite build tool:
 +
 ----
-kaitaistruct @ git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
+    kaitaistruct @ git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
```
generalmimon: Is this added indent actually important?

lang_python.adoc
```diff
@@ -114,30 +130,33 @@ https://setuptools.readthedocs.io/en/latest/[Setuptools] is the de-facto
...
+https://gitlab.com/KOLANICH/MFCTool.py/blob/master/setup.py[4],
+https://gitlab.com/KOLANICH/RDataParser.py/blob/master/setup.py[5].
 
 Here are
-https://github.com/KOLANICH/kaitaiStructCompile.py/blob/master/kaitaiStructCompile/config.schema.json[the
-docs] in JSON Schema format.
+https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.schemas/blob/master/kaitaiStructCompile/schemas/schemas/config.schema.json[the
+docs] in https://json-schema.org/specification.html[JSON Schema] format.
 
 Please remember that this lib is VERY UNSTABLE and very likely to be
```
generalmimon: I'd prefer not to use the "screaming case" to emphasize the text, in this case you'd probably want to use bold.

```suggestion
Please remember that this lib is **very unstable** and very likely to be
```

lang_python.adoc
```diff
@@ -114,30 +130,33 @@ https://setuptools.readthedocs.io/en/latest/[Setuptools] is the de-facto
...
-Please note that this lib is NOT available on
-https://pypi.python.org[pypi] and if you want to install it
-automatically, you should add
-`git+https://github.com/KOLANICH/kaitaiStructCompile.py` into
-`dependency_links` in the config passed to `setuptools.setup`.
+Please note that this lib is NOT available on https://pypi.python.org[pypi]
+and if you want to install it automatically, you should add
+`kaitaiStructCompile.setuptools @ https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools`
+into `setup_requires` into `options` section of your `setup.cfg` and into
+`requires` of `build_system` section of your `pyproject.toml`.
```
generalmimon: Uh, this is _way too many_ machine identifiers in one sentence, I've been lost a long time ago. It would be better to directly show the line `kaitaiStructCompile.setuptools @ https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools` in action, on the same place where it should be in practice (i.e. use a code block with a minimal samples of the resulting files after the additions).

lang_python.adoc
KOLANICH: thanks, fixed.

lang_python.adoc
KOLANICH: thanks, fixed.

lang_python.adoc
KOLANICH: thanks, fixed.

lang_python.adoc
KOLANICH: thanks, fixed. Found a neutral term `snapshot`.

lang_python.adoc
KOLANICH: thanks, fixed.

lang_python.adoc
KOLANICH: used the term `snapshot`

lang_python.adoc
```diff
@@ -13,98 +13,118 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
 [source,shell]
-kaitai-struct-compiler -t python my_spec.ksy
+kaitai-struct-compiler -t python [--python-package .] my_spec.ksy
 
 This normally generates `my_spec.py` parser module, which is compatible
-with both Python 2 and 3.
+with both Python 2 and 3. It is strongly recommended to use python 3
+because python 2 is EOL since 2020-01-01.
+
+`--python-package` optional CLI arg allows you to select a parent package
```
generalmimon: Can you please add the code block with the Python statement it gets translated into? And does it get inserted at all times, or only when `meta/imports` in the `.ksy` spec are used? I actually don't know what this option is used for, but _name used by one generated file to import symbols from another one_ seems related to _importing_ in KS terminology. And how does it relate to opaque types? IMHO all this info should be present here. 

I think people familiar with Python can better understand this CLI option when they see the Python equivalent, rather than when they have to read such a long paragraph of continuous text, which I didn't quite understand either. I've already suggested this in the second comment of https://github.com/kaitai-io/kaitai_struct_doc/pull/32#pullrequestreview-459605007. 

Every time when there's a opportunity to make the text shorter and more concise, e.g. explain the same thing with some code snippets or pictures imstead, it's a good idea to do it.

KOLANICH: I have added the example. But it hasn't made the text shorter, because some rationale to do so is still needed. It is obvious, but there are too many people for whom it is not.

lang_python.adoc
```diff
@@ -13,98 +13,138 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
-With stable version of Kaitai Struct compiler, one can use use following
+With stable version of Kaitai Struct compiler, one can use the following
 methods:
 
 * On Debian / Ubuntu, one can install it from apt repositories:
 
-** For Python 2.x:
-+
+** For Python 3.x:
+
```
generalmimon: You should add here the list continuation sign `+` here (and on a few other places), because right now the shell code block and the reference links below break the continuity of the list item "For Python 3.x:". Therefore you can see the sudden decrease of indentation where the `sudo apt-get` code block starts, when it should be horizontally aligned with the list item bullet instead. 

All paragraphs that belong to a particular list item are separated with the plus sign `+` instead of the blank line in Asciidoc, see [Asciidoctor quick reference](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#lists), the last example _Complex content in outline lists_. 

```suggestion
+
```

lang_python.adoc
KOLANICH: Done.

lang_python.adoc
KOLANICH: Done, thanks.

lang_python.adoc
```diff
@@ -13,98 +13,140 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
 `-t python` option:
 
 [source,shell]
-kaitai-struct-compiler -t python my_spec.ksy
+kaitai-struct-compiler -t python --python-package . my_spec.ksy
 
 This normally generates `my_spec.py` parser module, which is compatible
-with both Python 2 and 3.
+with both Python 2 and 3. It is strongly recommended to use python 3
+because python 2 is EOL since 2020-01-01. You can also install
```
generalmimon: Python should be written with the first uppercase letter (it's a name), writing it somehow else in a normal text looks unprofessionally.

```suggestion
with both Python 2 and 3. It is strongly recommended to use Python 3
because Python 2 is EOL since 2020-01-01. You can also install
```

lang_python.adoc
```diff
@@ -13,98 +13,140 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+`python` command.
+
+`--python-package` optional CLI arg allows you to select a parent package
+name used by one generated file to import symbols from another one.
+
+If you want to embed KSC-generated files into an own package, it is
+recommended to set the arg to `.`, which means "current package",
+rather than to the real  name of the subpackage, because setting it
+to a real name will cause breakage when you (or anyone else) want
+to reorganize or rename your package. It will cause KSC generate
```
generalmimon: A missing word:

```suggestion
to reorganize or rename your package. It will cause KSC to generate
```

lang_python.adoc
```diff
@@ -13,98 +13,140 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+(note the dot we have passed as an arg) instead of
+
+----
+from your_imported_spec import YourImportedSpec
+----
+.
+
+**Without setting this arg at all you will likely get an `ImportError`,
+if any of the specs imports another spec.** You can instead set this
+arg to the absolute path of the submodule, but it opens a can of worms.
```
generalmimon: > but it opens a can of worms.

I think it's better to avoid such expressive and vague phrases in the official documentation, and rather be specific what kind of problems it brings. Consider something like "but this brings problems with ..." or "but this brings problems when you ...".

KOLANICH: Thanks, fixed.

lang_python.adoc
KOLANICH: thanks, fixed.

lang_python.adoc
KOLANICH: Thanks, fixed.

lang_python.adoc
```diff
@@ -13,130 +13,167 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
-are welcome!]
-
-Please note that this lib is NOT available on
-https://pypi.org/[pypi] and if you want to install it
-automatically, you should add
-`git+https://github.com/KOLANICH/kaitaiStructCompile.py` into
-`dependency_links` in the config passed to `setuptools.setup`.
+https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.py/issues[Contributions are welcome!]
+
+Please note that this lib is NOT available on https://pypi.python.org[pypi]
```
generalmimon: Unnecessary "screaming case":

```suggestion
Please note that this lib is **not** available on https://pypi.python.org[pypi]
```

lang_python.adoc
```diff
@@ -13,130 +13,167 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
   (Python package manager):
 +
 [source,shell]
-pip install --upgrade --pre git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
+python3 -m pip install --upgrade --pre git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
+
 
 [[add-dependencies]]
-=== Adding dependencies
+=== Adding dependencies manually
```
generalmimon: I'd rather go for

```suggestion
=== Installing dependencies manually
```

..., because _adding_ is weird in this context. Why would someone want to _add_ dependencies? Which dependencies, i.e. dependencies of _what_? Where to add them? Installing is a much clearer term.

lang_python.adoc
KOLANICH: done

lang_python.adoc
KOLANICH: https://github.com/kaitai-io/kaitai_struct_doc/pull/32#discussion_r373444064

lang_python.adoc
```diff
@@ -13,130 +13,167 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+python3 -m pip install --upgrade kaitaistruct
 
 [[add-runtime-snapshot]]
 ==== Using latest (snapshot) version
 
-If you're using latest (unreleased AKA "unstable" AKA "snapshot") build
-of Kaitai Struct compiler, that will require latest runtime libraries as
-well:
+If you're using latest (unreleased AKA "bleeding edge" AKA "unstable"
+AKA "snapshot" AKA "nightly") build of Kaitai Struct compiler, that will
```
generalmimon: I don't really like using the _AKA_ abbreviation that many times as the delimiter of the alternatives, what about using commas and _or_ before the last one? Another thing that I'm not glad to see is the a.k.a. abbreviation written without dots with all letters uppercase. I remember that when I first encountered it in the KS docs, I had to search it by Google what it's supposed to mean (if it's some special technical thing), because I've never seen it written other than like `a.k.a.` before. The uppercase acronyms are generally more often used for artificial names like JSON, XML, JPEG, ..., not for common "language constructs" like etc. (_et cetera_), e.g. (_exempli gratia_), ca. (_circa_), and so on.

The best option is of course not to use the abbreviations at all to avoid confusion.

```suggestion
If you're using latest (unreleased, also known as "bleeding edge", "unstable",
"snapshot" or "nightly") build of Kaitai Struct compiler, that will
```

lang_python.adoc
generalmimon: Yeah, I have read it and knew you'd bring that, but it seems that the idea hasn't gone far. I looked into the other language-specific notes and none of them actually contain anything like "adding runtime library" or "adding dependencies". [Java](https://doc.kaitai.io/lang_java.html), [JavaScript](https://doc.kaitai.io/lang_javascript.html) and [PHP](https://doc.kaitai.io/lang_php.html) notes conversely use the term _install_ for runtime library, and IMHO one does exactly the same activity with dependencies.

As for the anchor, I'd say feel free to change it according to the heading.

lang_python.adoc
generalmimon: > It is obvious, but there are too many people for whom it is not.

It may be obvious to you, but it wasn't to me until you added the example. Now it's much better than before 👍

lang_python.adoc
KOLANICH: The traditional spelling is without dots:

[https://writingexplained.org/aka-or-aka](https://writingexplained.org/aka-or-aka)

https://dictionary.cambridge.org/dictionary/english/aka

https://lurkmore.to/Aka

generalmimon: OK, thanks for the stats. But I don't think that the uppercase variant is too frequent, the letter case is my main concern. And it doesn't change the fact that it's generally better to use the full variant whenever possible.

KOLANICH: `aka` is a kind of a meme and is not really used in its expanded variant.

lang_python.adoc
```diff
@@ -13,130 +13,167 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+independent fork of your lib.
+
+It is also recommended to put all the KSC-generated code into an
+own subpackage, in example `kaitai`, so, if your build process
+regenerates the files on every build, you could just .gitignore that dir.
 
-TODO: document `--python-package`
 
 [[add-runtime]]
 === Adding runtime library
```
generalmimon:
```suggestion
=== Installing runtime library
```

... and please also update the anchor accordingly.

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
 
 [source,shell]
-kaitai-struct-compiler -t python my_spec.ksy
+kaitai-struct-compiler -t python --python-package . my_spec.ksy
 
 This normally generates `my_spec.py` parser module, which is compatible
-with both Python 2 and 3.
+with both Python 2 and 3. It is strongly recommended to use Python 3
+because Python 2 is EOL since 2020-01-01. You can also install
+`python-is-python3` on Ubuntu focal to make `python3` be used via
```
generalmimon: What is _focal_, or _Ubuntu focal_?

dgelessus: "Focal Fossa" (or "focal" for short) is the codename/internal name of Ubuntu 20.04.

I'm not sure if the `python-is-python3` part is relevant enough to be included here at all - not everyone uses Ubuntu, not every Ubuntu user is on 20.04 yet, and not everyone on 20.04 wants to (or is able to) use `python-is-python3`. I think it's enough to say that Python 2 is no longer supported and Python 3 should be preferred, and let users figure out how to handle Python versions on their system.

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+`python` command.
+
+`--python-package` optional CLI arg allows you to select a parent package
+name used by one generated file to import symbols from another one.
+
+If you **don't intend to embed** the KSC-generated files and just want to
+play a bit in IPython/Jupyter shell in the same dir, **remove** this arg,
```
dgelessus: I would use either "REPL" or "Python shell" here, because this also applies when using the default interactive Python shell, not just IPython.

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+otherwise you get an `ImportError` .
+
+If you **intend to embed** the compilation result into an own package,
+it is recommended **to set the arg to `.`**, which means "current package",
+rather than to the real  name of the subpackage, because setting it
+to a real name will cause breakage when you (or anyone else) want
+to reorganize or rename your package. It will cause KSC to generate
```
dgelessus: I would shorten this paragraph to only talk about `--python-package .` and leave out the discussion about `--python-package absolute.package.name`. That way it's less confusing for people who don't already know what `--python-package` does.

I don't know of any cases where `--python-package .` does not work and you have to use `--python-package absolute.package.name` instead. I would either remove the discussion of the absolute name variant completely, or move it somewhere to the bottom.

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+----
+from your_imported_spec import YourImportedSpec
+----
+.
+
+**Without setting this arg at all you will likely get an `ImportError`,
+if any of the specs imports another spec.** You can instead set this
+arg to the absolute path of the submodule, but it violates DRY principle
+and will only result in additional work when refactoring or making an 
+independent fork of your lib.
```
dgelessus: As above, I don't think it's worth discussing the absolute path variant of `--python-package` in detail (especially not in two places).

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+
+It is also recommended to put all the KSC-generated code into an
+own subpackage, in example `kaitai`, so, if your build process
+regenerates the files on every build, you could just .gitignore that dir.
```
dgelessus: I would split this into two sentences:
```suggestion
It is also recommended to put all the KSC-generated code into an
own subpackage, in example `kaitai`. If your build process
regenerates the files on every build, you could just .gitignore that dir.
```

lang_python.adoc
```diff
@@ -13,130 +13,171 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
-TODO: document `--python-package`
 
-[[add-runtime]]
-=== Adding runtime library
+[[install-runtime]]
+=== Installing runtime library
 
 Code generated by Kaitai Struct compiler will require additional runtime
-library (https://pypi.org/project/kaitaistruct/[kaitaistruct at PyPi],
+library (https://pypi.org/project/kaitaistruct/[image:https://img.shields.io/pypi/v/kaitaistruct.svg[kaitaistruct at PyPi]],
```
dgelessus: Minor thing: it's called "PyPI", not "PyPi".

```suggestion
library (https://pypi.org/project/kaitaistruct/[image:https://img.shields.io/pypi/v/kaitaistruct.svg[kaitaistruct at PyPI]],
```

lang_python.adoc
KOLANICH: The problem is that in PEP 8 (IMHO, in some places it is a complete piece of shit) it is written it is recommended to use absolute imports. After it it is written that relative imports are also allowed. But some people are PEP8 fanatics ...

lang_python.adoc
KOLANICH: I don't think that removal of the conjunction is beneficial. IMHO the removal harms the meaning.

lang_python.adoc
KOLANICH: https://github.com/kaitai-io/kaitai_struct_doc/pull/32#discussion_r464110839

lang_python.adoc
KOLANICH: Thanks, fixed.

lang_python.adoc
KOLANICH: Thanks, fixed.

lang_python.adoc
dgelessus: It's only a recommendation, so it probably won't cause problems with automated code style checkers (which shouldn't be applied to generated code anyway). Besides, the Python code generated by KS already violates other PEP 8 guidelines, such as those about how to group imports, so I think nobody expects KS to generate strictly PEP 8-compliant Python code.

In any case, I still think it's not important enough to mention this early in the section.

lang_python.adoc
dgelessus: I think it's the opposite actually. The conjunction implies that .gitignoring the generated files is the main/only reason to put them in a separate package, which I don't think is the case. I would say that the main reason to put them in a separate package is keeping the code well organized and avoiding naming conflicts between generated and non-generated code (so that KSC can't overwrite non-generated code by accident).

lang_python.adoc
KOLANICH: Shortened the section and removed all the unneeded details.

lang_python.adoc
```diff
@@ -13,130 +13,155 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
+  adding a https://www.python.org/dev/peps/pep-0508/[PEP 508]
+  dependency specifier into the config of your favorite build tool:
 +
 ----
 kaitaistruct @ git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git
 ----
 +
-and then run `pip install -r requirements.txt` to update all your
-dependencies.
+.
```
generalmimon: I'd really rather remove this dot, as a lone dot on a separate line looks quite weird when you display the rendered Asciidoctor output. On top of that, you aren't adding the full stop after the code block like this anywhere else in the document.

lang_python.adoc
```diff
@@ -13,130 +13,155 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
-setuptools], allowing you to automate compilation of `*.ksy` files as the
-part of python module build process. The library provides a mostly
-declarative syntax to facilitate installation: you add a parameter
-`kaitai` to your `setup.py` where you enumerate files that you want
-compiled, output directory, compilation flags, and source code
-post-compilation transformations. Here are some examples:
-https://github.com/KOLANICH/NTMDTRead/blob/master/setup.py[1],
-https://github.com/KOLANICH/SpecprParser.py/blob/master/setup.py[2].
+https://gitlab.com/KOLANICH has developed
+https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools[an extension to setuptools] ( https://github.com/kaitaiStructCompile/kaitaiStructCompile.setuptools[GitHub mirror is also present]), allowing you to automate compilation of `*.ksy` files as the
```
generalmimon: An additional space after the opening bracket here:

```suggestion
https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools[an extension to setuptools] (https://github.com/kaitaiStructCompile/kaitaiStructCompile.setuptools[GitHub mirror is also present]), allowing you to automate compilation of `*.ksy` files as the
```

You could also split this extra long line into several shorter ones to make the source more readable:

```suggestion
https://gitlab.com/kaitaiStructCompile.py/kaitaiStructCompile.setuptools[an extension to setuptools]
(https://github.com/kaitaiStructCompile/kaitaiStructCompile.setuptools[GitHub mirror is also present]),
allowing you to automate compilation of `*.ksy` files as the
```

lang_python.adoc
KOLANICH: Done.

lang_python.adoc
KOLANICH: Done.

lang_python.adoc
```diff
@@ -13,130 +13,155 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
...
 
-TODO: document `--python-package`
+`--python-package` is an optional CLI arg allows you to select a parent package
+name used by one generated file to import symbols from another one.
 
-[[add-runtime]]
-=== Adding runtime library
+It is recommended to just follow the algorithm:
+
+* If you don't want to get an `ImportError`, determine if you need `--python-package`:
```
generalmimon: This list item (which brings an additional level of nesting) is IMHO completely unnecessary. Who would want to get an `ImportError`? Everybody wants everything to work seamlessly on the first try without errors. Rather skip directly to the point. You can marginally mention at the end of the sentence somewhere that when you don't do this, you'll receive an `ImportError`, but it's not the main point.

KOLANICH: Removed.

generalmimon (approved):
Looks good to me, thanks a lot for this massive effort, and especially for your awesome attitude to the review comments! The PR has improved a lot since your initial proposal, and you managed to address most of the suggestions in a single day. Great job!

What amazes me the most is that though I may be really annoying with the big pile of minor corrections, you never lose your temper and you even thank me for most of my findings. Thank you again for keeping the good mood here!

lang_python.adoc
KOLANICH: Fixed.

lang_python.adoc
generalmimon: > `aka` is a kind of a meme and is not really used in its expanded variant.

I don't really think so, see this random page on Wikipedia for example: https://en.wikipedia.org/wiki/Splitting_maul

> A **splitting maul** also known as a _block buster_, _block splitter_, _chop and maul_, _sledge axe_, _go-devil_ or _hamaxe_ is a heavy, long-handled axe used for splitting a piece of wood along its grain.

Just imagine how it would look if Wikipedia would use _aka_ between every two alternatives instead of the comma (`,`).

KOLANICH: Usage of memes is a higly informal style, it is not the style accepted on wikipedia. It is more accepted in dramatica and other places rather than on wikipedia.

generalmimon: > Usage of memes is a higly informal style, it is not the style accepted on wikipedia. It is more accepted in dramatica and other places rather than on wikipedia.

If this style is not accepted by Wikipedia, why it should be accepted by the official Kaitai Struct documentation? Are the KS docs somewhat more inferior compared to Wikipedia? 

I think we should better try to make the docs look more professionally. _Open source_ doesn't imply _less significant_ or _more informal_. Our official documentation should not look like a chat.

On top of that, Kaitai Struct is used by users from around the world, and many of them don't have English as a native language. Using informal language makes it harder for these people to understand, because they're usually not familiar with the colloquial English used by native speakers for everyday communication.

KOLANICH: >I think we should better try to make the docs look more professionally.

Are you trying to raise funding from some sponsor, who demands to remove it? If not, why should we?

>Are the KS docs somewhat more inferior compared to Wikipedia?

No, it is Wikipedia that inferior to less formal encyclopædias in some sense. Some facts cannot go to wikipedia because of WP rules about what WP is for and what is it not for. You probably shouldn't apply WP rules to the rest of the world.

>Our official documentation should not look like a chat.

It doesn't.

>On top of that, Kaitai Struct is used by users from around the world, and many of them don't have English as a native language.

The first time I saw this abbreviation, it was in a magazine (printed on paper) written in Russian around 2004. Though I have never met the abbreviation before, I quickly derived from the context what it means.

generalmimon: > Are you trying to raise funding from some sponsor, who demands to remove it? If not, why should we?

You don't need a multi-million dollar budget to change a single sentence to comply with the formal language and style.

Why should we do it? To make the docs more readable, understandable and clear, to make the docs look reputable, and also not to discourage potential users by the bad condition or language level of the docs. 

> No, it is Wikipedia that inferior to less formal encyclopædias in some sense. Some facts cannot go to wikipedia because (...)

I'm sorry, did I say a word about _facts_? I've only used Wikipedia as an example site from which we can still learn a few things regarding to language and style of the text in my opinion. Wikipedia content may be inaccurate or incorrect, but I don't see any reason why we should discuss it here. That's a whole different matter.

If you consider the language and style level on Wikipedia still too low in some sense, please find a more reputable site which uses literally _AKA_ or _aka_ and share a link to it.

> > Our official documentation should not look like a chat.
> 
> It doesn't.

Using _aka_ four times in a row seems more appropriate rather for an informal chat than for any official documentation. It's like when I write _"I need to buy apples and bread and ham and milk and meat."_. You don't need to be a linguist to notice that something is wrong.

> The first time I saw this abbreviation, it was in a magazine (printed on paper) written in Russian around 2004. Though I have never met the abbreviation before, I quickly derived from the context what it means.

You wouldn't have to derive anything if the magazine used the expanded "also known as" in Russian instead. Besides, even if the particular _aka_ word was understandable by 100% of our audience, it lowers the overall language style and feeling of reputability of the docs. Repeating it multiple times in a row instead of using a regular comma makes it even worse.

KOLANICH: >it lowers the overall language style

It is actually a good thing.

>feeling of reputability of the docs

We are neither a parliament, nor an army headquaters, nor a court. We are just a group of people working on a cool tool. And this style brings more reputability and trust then trying to look more serious than we are.

generalmimon: > > it lowers the overall language style
>
> It is actually a good thing.

> And this style brings more reputability and trust then trying to look more serious than we are.

I really doubt that, but I can see that I'm not going to move the mountain. 
I hope you're right, as I'm not willing to waste any more hours in this discussion. Let's focus on some stuff that matters.

PR #34, opened 2020-02-22
Syntax highlighting: Replace CodeRay with Pygments

generalmimon:
I suggest switching from [CodeRay](http://coderay.rubychan.de/) to [Pygments](https://pygments.org/). CodeRay doesn't support YAML much (for instance, see [this YAML test](http://coderay.rubychan.de/rays/7425) or [KSY sample](http://coderay.rubychan.de/rays/9274)) and doesn't really seem to be in active development. It supports only 22 languages (for example C#, Visual Basic, Swift, TypeScript, Perl, Kotlin, Rust, Scala and Haskell are missing).

[Pygments](https://pygments.org/) is a popular syntax highlighter. It supports almost 250 languages and a number of styles (here's a [online demo where you can play with it](https://pygments.org/demo/)). But the best thing is the support for [writing custom lexers](https://pygments.org/docs/lexerdevelopment/). This is the reason why I chose it: I would like to write a custom lexer for KSY, which would highlight the KS expression language as well. This is a prerequisite, but this change will immediately improve the readability of all code samples as well.

I installed the [pygments.rb gem](https://rubygems.org/gems/pygments.rb/) as suggested by the Asciidoctor docs and switched the `source-highlighter` to `pygments`. It worked, but the YAML highlighting was still somehow broken, while e.g. C code highlighting worked fine. I found that the pygments.rb is using some old version of Pygments, which had this issue: https://bitbucket.org/birkenfeld/pygments-main/pull-requests/762. So I install the latest Pygments with `pip install` and then created `pygments_init.rb`, which makes Asciidoctor to use the custom installation (see [Using a custom Pygments installation in Asciidoctor docs](https://asciidoctor.org/docs/user-manual/#using-a-custom-pygments-installation)).

If you don't like the `default` Pygments style, feel free to choose something nicer from https://pygments.org/demo/#style 🙂

### DIFF
--- a/pygments_init.rb
+++ b/pygments_init.rb
@@ -1,0 +1,4 @@
+require 'pygments'
+
+# use a custom Pygments installation (directory that contains pygmentize)
+Pygments.start `dirname "$(which pygmentize)"`
--- a/pygments_init.rb
+++ b/pygments_init.rb
@@ -1,4 +1,4 @@
 require 'pygments'
 
 # use a custom Pygments installation (directory that contains pygmentize)
-Pygments.start `dirname "$(which pygmentize)"`
+Pygments.start `which pygmentize`
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -2,7 +2,7 @@
 Kaitai Project
 v0.8
 :toc: left
-:source-highlighter: pygments
+:numbered:
 
 == Introduction
 
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -2,8 +2,7 @@
 Kaitai Project
 v0.8
 :toc: left
-:source-highlighter: coderay
-:numbered:
+:source-highlighter: pygments
 
 == Introduction
 
@@ -2310,7 +2309,7 @@ inside strings literals. So, trying something like that would fail in
 some parsers (namely, it *will* fail using "desktop"
 kaitai-struct-compiler running under JVM):
 
-[source,java]
+[source,yaml]
 ----
 instances:
   foo:
@@ -2319,7 +2318,7 @@ instances:
 
 To ensure maximum compatibility, put quotes around such strings, i.e:
 
-[source,java]
+[source,yaml]
 ----
 instances:
   foo:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ cache:
   - apt
 
 before_install:
-  - gem install asciidoctor coderay
+  - gem install asciidoctor pygments.rb
+  - pip install Pygments
 
 script:
   - make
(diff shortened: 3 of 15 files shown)


user_guide.adoc
```diff
@@ -2,8 +2,7 @@
 Kaitai Project
 v0.8
 :toc: left
-:source-highlighter: coderay
-:numbered:
+:source-highlighter: pygments
```
GreyCat: If this is not needed in all files, then probably we should not add it here. If it is actually required, then we should replace `:source-highlighter: coderay` in all files with this, not just delete it.

user_guide.adoc
```diff
@@ -2,8 +2,7 @@
 Kaitai Project
 v0.8
 :toc: left
-:source-highlighter: coderay
-:numbered:
```
GreyCat, line 6: Why have you removed numberings of chapters?

pygments_init.rb
```diff
@@ -0,0 +1,4 @@
+require 'pygments'
+
+# use a custom Pygments installation (directory that contains pygmentize)
+Pygments.start `dirname "$(which pygmentize)"`
```
GreyCat: I'm not sure if that actually works as intended. On my system, pygmentize was coming from `python3-pygments` .deb package, thus it points to `/usr/bin/pygmentize`, thus dirname of that will be `/usr/bin`. I assume what is really supposed to be found in this situation is `/usr/lib/python3/dist-packages/pygments`?

Also, in theory this can be written in Ruby without a shell call.

user_guide.adoc
generalmimon: Good catch, it's my mistake, sorry. Of course it's not needed, because I set the `source-highlighter` in the Makefile.

pygments_init.rb
generalmimon: > On my system, pygmentize was coming from `python3-pygments` .deb package, thus it points to `/usr/bin/pygmentize`

Yes, I think it _should_ work fine. Executing `which pygmentize` returns `/c/Python27/Scripts/pygmentize` for me (I'm using Git Bash on Windows). And I follow the comment's suggestion (I took this approach from the Asciidoctor docs: [Using a custom Pygments installation](https://asciidoctor.org/docs/user-manual/#using-a-custom-pygments-installation), "directory that contains pygmentize", so I used `dirname`. But I've just discovered that it works even without it, so I'll throw the `dirname` out.

> Also, in theory this can be written in Ruby without a shell call.

I didn't know that, if so, it would be probably better. My knowledge of Ruby is limited 😄  We can replace it with [something like this](https://stackoverflow.com/a/5471032/) if you prefer.

user_guide.adoc
generalmimon: My bad, fixed.

pygments_init.rb
GreyCat: Ok, I get the idea, let's leave it as is now. Thanks!

PR #35, opened 2020-05-11
Fix `_index` examples, add behavior of keys in repetition

### LINKED ISSUE #436: Referencing _index in the pos attribute of a repeated instance produces: undefined local variable or method `i'
Here's about the simplest example of the problem I could come up with. This is almost verbatim the second example in [7.9. Repetition index](http://doc.kaitai.io/user_guide.html#repeat-index). I've just made things shorter by using `u1` instead of `u4`, and removing the `file_names` item.

`files.ksy`:

    meta:
      id: files
    seq:
      - id: num_files
        type: u1
      - id: len_files
        type: u1
        repeat: expr
        repeat-expr: num_files
      - id: ofs_files
        type: u1
        repeat: expr
        repeat-expr: num_files
    instances:
      file_bodies:
        pos: ofs_files[_index]
        size: len_files[_index]
        repeat: expr
        repeat-expr: num_files

And we want to use this on `files.bin`:

    00000000: 02 05 04 0a 13 00 00 00 00 00 61 62 63 64 65 00 | ..........abcde.
    00000010: 00 00 00 77 78 79 7a                            | ...wxyz

When you run `ksv files.bin files.ksy`, everything is initially fine. But when you try to expand the `file_bodies` node, you get:

    files.rb:42:in `file_bodies': undefined local variable or method `i' for Files(num_files=2 len_files=[5, 4] ofs_files=[10, 19]):Files (NameError)

If you compile to Ruby, it's obvious what's wrong:

      def file_bodies
        return @file_bodies unless @file_bodies.nil?
        _pos = @_io.pos
        @_io.seek(ofs_files[i])
        @file_bodies = Array.new(num_files)
        (num_files).times { |i|
          @file_bodies[i] = @_io.read_bytes(len_files[i])
        }
        @_io.seek(_pos)
        @file_bodies
      end

It seems `ksc` interprets the `pos` attribute as outside of the repetition loop, when it should be inside. I've only looked at the Ruby, Java and CSharp output, but it seems to be doing this in all target languages.

generalmimon:
Fixes https://github.com/kaitai-io/kaitai_struct/issues/436

Related:

- https://github.com/kaitai-io/kaitai_struct/issues/14
- https://github.com/kaitai-io/kaitai_struct/issues/145
- https://github.com/kaitai-io/kaitai_struct/issues/281
- https://github.com/kaitai-io/kaitai_struct/issues/317
- https://github.com/kaitai-io/kaitai_struct/issues/376
- https://github.com/kaitai-io/kaitai_struct/issues/606
- https://github.com/kaitai-io/kaitai_struct/issues/722

I found that writing documentation isn't exactly a simple task and it takes a lot of time to figure out the right phrases, but it has to be done. This section is missing in the docs probably for years (among others). KS users often encounter various issues resulting from not understanding which keys apply to what (just look at the issues I linked, and the [recent conversation on Gitter](https://gitter.im/kaitai_struct/Lobby?at=5eb5db5522f9c45c2a7b69ac) reminded me that it persists).

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -683,6 +683,8 @@ KS supports all these types of repetitions. In all cases, it will create
 a resizable array (or nearest equivalent available in target language)
 and populate it with elements.
 
+
+[[repeat-eos]]
 ==== Repeat until end of stream
 
 This is the simplest kind of repetition, done by specifying
@@ -1676,6 +1678,7 @@ For `comment`, we just made it to have size up until the end of
 stream. Given that we've limited it to the substream in the first
 place, this means exactly what we wanted.
 
+[[repeat-until-size-limit]]
 === Repeating until total size reaches limit
 
 The same technique might be useful for repetitions as well. If you
@@ -1709,12 +1712,13 @@ repeat until the end of that stream.
 However, note that one's naïve approach might not work:
 
 * When we're dealing with an array of elements, `size` will refer to
-the size of one particular element of the array.
-* Any repetition (and this includes `repeat: eos`) uses current
-  object's IO stream. Substreams are created individually for every
-  object inside the loop.
+  the size of one particular element of the array.
+* Any repetition (and this includes `repeat: eos`) parses the elements
+  using the current IO stream. If `size` is specified, substreams
+  are created individually for each object inside the loop.
 
-So this is wrong:
+So this is *wrong* (`total_len` determines the size
+of each individual `entry` substream here):
 
 [source,yaml]
 ----
@@ -1726,6 +1730,8 @@ seq:
     size: total_len
     repeat: eos
 ----
+
+For more information, see <<keys-repeated>>.
 
 =====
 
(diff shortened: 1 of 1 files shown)


generalmimon:
@GreyCat @KOLANICH @dgelessus @jchv @cugu Any feedback?

jchv:
I think it LGTM after a quick read-over, as long as I understand correctly how these functions of Kaitai Struct work.

dgelessus (approved):
No specific feedback from me either - all looks good to me.

Thank you for taking the time to write this up! It's good to have less obvious behavior like this documented properly, along with workarounds for when the default behavior doesn't do what you need.

generalmimon:
@GreyCat Could you please also add your review?

generalmimon:
I'm merging this, because I think that it's better to have this merged as it can help some people, than to let this PR decay among the open pull requests here, even if it's not perfect. At least I tried to get some criticism 😆

PR #39, opened 2020-07-04
Web IDE: link to feature list and devel version

generalmimon:
See https://github.com/kaitai-io/kaitai_struct_webide/issues/116#issuecomment-644395058

I'm tired of telling people all the time that we have also the [devel Web IDE](https://ide.kaitai.io/devel/), where they can use all latest features. Why not save our time by writing it to the User Guide instead? 😃


generalmimon:
@dgelessus left a thumb, so I guess it can useful for some people, merging.

PR #41, opened 2020-08-04
Document exact Python versions supported by KSC and runtime

### LINKED ISSUE #695: Supporting old Python versions
This topic came up in multiple recent issues (#675, kaitai-io/kaitai_struct_python_runtime#25) - which old Python versions should we support for how long?

## Which versions does KS currently support?

kaitai_struct_python_runtime is compatible with *a lot* of Python versions, because it doesn't use many advanced Python features. The minimum required version seems to be Python 2.6, which added support for byte-string literals (`b"..."`) and the `io` module. Python 3 versions before 3.3 are not supported: the runtime uses [setup.cfg-based declarative configuration](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files), which was added in setuptools 30.3.0, and setuptools 30.0.0 [dropped support for Python 3.2](https://setuptools.readthedocs.io/en/latest/history.html#v30-0-0). (This just applies to the package metadata though - the actual source code might even be compatible with Python 3.0-3.2.)

The Python code generated by ksc is also compatible with Python 2.6 and later (and possibly even older, if it wasn't limited by kaitai_struct_python_runtime's requirements). There is one exception: if the spec uses enums, the generated code uses the `enum` module, which was only added to the standard library in Python 3.4 (and doesn't exist in Python 2). However, the [enum34](https://pypi.org/project/enum34/) library provides a backport of the `enum` module to older Python versions, down to 3.1 and 2.4.

In short, we currently support Python 3.4 and later, and if enum34 is installed also 3.3, 2.7 and 2.6.

Note: [the CI](https://ci.kaitai.io/) only tests on Python 2.7 and 3.6, so it's possible that there are subtle bugs on versions older than that.

## Which Python versions are still supported, and which are EOL?

The question of which Python versions are "supported" or "EOL" is a little complicated.

Most of the time, we're talking about what is still supported by the CPython developers. (CPython is the reference implementation of Python and is used by the vast majority of Python users.) By that definition, [Python 2.6 has been EOL since 2013-10-29](https://www.python.org/dev/peps/pep-0361/#release-lifespan), [Python 3.3 since 2017-09-29](https://www.python.org/dev/peps/pep-0398/#lifespan), [Python 3.4 since 2019-03-18](https://www.python.org/dev/peps/pep-0429/#release-schedule), and [Python 2.7 since 2020-01-01](https://www.python.org/dev/peps/pep-0373/#maintenance-releases) (although the final Python 2.7 version will be released in "mid-April, 2020").

However, this is only what the CPython devs themselves support - there are third parties that provide longer support for CPython versions that the upstream devs no longer support. Most notably this includes Linux distributors who provide long-term support releases of their distros (e. g. Ubuntu, RHEL, CentOS) - they support the Python versions that shipped with each release for the entire support duration of that OS release. This is especially relevant for Python 2.7, which is included in Ubuntu 18.04 LTS (supported until 2023-04, and paid support until 2028-04) and RHEL and CentOS 8 (supported until 2024-05, security fixes until 2029).

PyPy, which is perhaps the biggest alternative implementation of Python, plans to [support its Python 2 version "forever"](https://pypy.readthedocs.io/en/latest/faq.html#how-long-will-pypy-support-python2).

## Which versions should we support in the future?

I think this should depend on what the Kaitai Struct userbase uses - unfortunately I don't think we have a lot of statistics for that. According to [PyPI stats for the kaitaistruct package](https://pypistats.org/packages/kaitaistruct), the majority of users use Python 3.6 and 3.7. Usage of 3.8, 3.5 and 2.7 is very small, and other versions are basically nonexistant. I don't know how reliable these statistics are though - PyPI download statistics are [known to be inaccurate](https://packaging.python.org/guides/analyzing-pypi-package-downloads/#background).

IMHO it would be safe to drop support for Python 2.6, 3.3 and probably also 3.4 - these versions are relatively old by now, and all of them have compatible newer versions that users can upgrade to. Python 2.7 is a different story - although it's technically no longer supported by the CPython devs, it's still used a lot, because migration from Python 2 to 3 can be difficult.

dgelessus:
Fixes kaitai-io/kaitai_struct#695 (the small remaining part of it).

I've moved the Python version compatibility info into a small separate section at the beginning, because it applies to both the compiler-generated code and the runtime library.

### DIFF
--- a/lang_python.adoc
+++ b/lang_python.adoc
@@ -6,6 +6,14 @@ v0.9
 [[overview]]
 == Overview
 
+[[supported-python-versions]]
+=== Supported Python versions
+
+Kaitai Struct supports Python 3.4 and later as well as Python 2.7.
+It is strongly recommended to use Python 3.5 or newer,
+because Python 3.4 and 2.7 have reached end-of-life
+and no longer receive official support from the Python developers.
+
 [[ksc]]
 === Kaitai Struct compiler invocation
 
@@ -15,9 +23,7 @@ To generate Python modules, Kaitai Struct compiler should be invoked with
 [source,shell]
 kaitai-struct-compiler -t python [--python-package .] my_spec.ksy
 
-This normally generates `my_spec.py` parser module, which is compatible
-with both Python 2 and 3. It is strongly recommended to use Python 3
-because Python 2 is EOL since 2020-01-01.
+This normally generates `my_spec.py` parser module.
 
 `--python-package` is an optional CLI arg allows you to select a parent package
 name used by one generated file to import symbols from another one.


generalmimon (approved):
@dgelessus Thanks!

PR #43, opened 2020-09-17
Add little-endian bit integers to the User Guide

generalmimon:
Documenting [little-endian bit integers](https://github.com/kaitai-io/kaitai_struct/issues/155#issuecomment-613733627) feature in anticipation of 0.9 release...

Tip: it turns out that GitHub can show rich rendered Asciidoc diffs, which make it easy to read. Here's the [rich diff 
of `user_guide.adoc` from this PR](https://github.com/kaitai-io/kaitai_struct_doc/pull/43/files?short_path=837cfbf#diff-837cfbf61fa96ac29d27628da1629b37). 

Cc @dgelessus, @tempelmann

_I fell in love with [box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character) :-)_

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -1084,7 +1084,7 @@ instances:
     value:  packed_1 & 0b00001111
 ----
 
-However, Kaitai Struct offers a better way to do it - using
+However, Kaitai Struct offers a better way to do it — using
 *bit-sized integers*.
 
 [[bit-ints-be]]
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -1056,8 +1056,6 @@ to enforce the type of the expression, see <<typecast,typecasting>>.
 
 === Bit-sized integers
 
-IMPORTANT: Feature available since v0.6.
-
 Quite a few protocols and file formats, especially those who aim to
 conserve space, pack multiple integers into same byte, using integer
 sizes less than 8 bits. For example, IPv4 packet starts with a byte
@@ -1071,7 +1069,30 @@ vvvvllll
   └───── version
 ....
 
-Here's how it can be parsed with KS:
+It's possible to unpack bit-packed integers using old-school
+methods with bitwise operations in value instances:
+
+[source,yaml]
+----
+seq:
+  - id: packed_1
+    type: u1
+instances:
+  version:
+    value: (packed_1 & 0b11110000) >> 4
+  len_header:
+    value:  packed_1 & 0b00001111
+----
+
+However, Kaitai Struct offers a better way to do it - using
+*bit-sized integers*.
+
+[[bit-ints-be]]
+==== Big-endian order
+
+IMPORTANT: Feature available since v0.6.
+
+Here's how the above IPv4 example can be parsed with KS:
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


ams-tschoening (approved):
Looks great, especially ASCII-art helps a lot.

user_guide.adoc
```diff
@@ -1082,17 +1103,44 @@ seq:
...
+....
+
+`x[0]` is the first byte of the stream, and the numbers 7-0 on the line
+below indicate the invididual bits of this byte (listed from MSB `7` to LSB `0`).
 
-Using `type: bX` (where X is a number of bits to read) is very
+The value of `version` can be retrieved as `0b{v3}{v2}{v1}{v0}`
+__(`0b...` is the binary integer literal as present in many programming
+languages, and `{v3}` is the value `0` or `1` of the corresponding bit)__,
+as well as `len_header` value can be written as `0b{l3}{l2}{l1}{l0}`.
```
dgelessus: Using the lowercase letter `l` right next to numbers can be a bit confusing, because depending on the font the letter `l` can look very similar to the number `1`. I would use a different letter here if possible - maybe `h` for header?

user_guide.adoc
```diff
@@ -1082,17 +1103,44 @@ seq:
...
+
+As you can see in the bit layout, `version` occupies https://en.wikipedia.org/wiki/Nibble#Low_and_high_nibbles[high nibble]
+of the byte `x[0]` and `len_header` occupies low nibble.
```
dgelessus: I'm not sure if it makes much sense to introduce the term "nibble" here, because it's not used anywhere else in this part of the docs. I would write "high 4 bits" and "low 4 bits" instead, so that it's a bit less confusing for people who don't already know what a nibble is. Or possibly remove this sentence completely - the bit layout is already explained well by the diagram above.

user_guide.adoc
```diff
@@ -1082,17 +1103,44 @@ seq:
...
+Using `type: bX` (where `X` is a number of bits to read) is very
 versatile and can be used to read byte-unaligned data. A more complex
 example of packing, where value spans two bytes:
 
 ....
-76543210 76543210
-aaaaabbb bbbbbbcc
+               x[0]                              x[1]
+  7   6   5   4   3   2   1   0     7   6   5   4   3   2   1   0
+  a4  a3  a2  a1  a0  b8  b7  b6    b5  b4  b3  b2  b1  b0  c1  c0
```
dgelessus, line 1138: The names `b8`, `b7`, etc. might be confusing, because those are also the names of bit-sized integer types. I would have suggested `xyz` instead of `abc`, but `x` is already taken by `x[0]`/`x[1]` - maybe write something like `data[0]`/`data[1]` instead, so that `xyz` can be used for the bit field names?

dgelessus (commented):
A general comment: I think it would be better if the docs would recommend setting `bit-endian` explicitly for both BE and LE cases. The problem is that `bit-endian: be` is the default *regardless of `endian`*. In the BE case that works fine, but in the LE case that default is nearly always wrong (so far I haven't seen a LE format where `bit-endian: be` would be the right choice). So I would avoid getting people into the habit of relying on the default `bit-endian` (when speccing a BE format) and then getting bitten by the bad default later (when speccing a LE format), and instead prefer explicitly setting the correct `bit-endian` in all cases.

The exception is of course when writing specs that need to be compatible with older compiler versions for whatever reason. In that case you can't set `bit-endian: be` explicitly and have to rely on the default (or rather the only possible behavior before 0.9). But I would explain that in the dedicated section on `bit-endian` at the bottom, because once 0.9 is released I would expect most people to target 0.9 and not care about older versions (in most cases there's no reason to stay compatible with older release versions of the compiler). This way it's also less confusing to someone who's reading these docs for the first time and doesn't already know about the backwards compatibility story - it's not obvious why `bit-endian: be` is the default and only `bit-endian: le` needs to be set explicitly, compared to regular `endian` which always needs to be set explicitly and has no default.

user_guide.adoc
```diff
@@ -1167,25 +1243,181 @@ i.e. two least significant bits of the first byte would be lost and
...
+members.
+
+This is because each byte in a *big-endian* bit field is gradually "filled"
+with members from the most significant bit (7) to the least significant (0),
+and if the current byte is filled up to LSB, the parsing continues on
+MSB of next byte. It follows that `b` really can't be represented
+with a single attribute using this order, because `c` and `a` are standing
+in the way.
+
+Little-endian bit fields use the reversed parsing direction: bytes are filled
```
KOLANICH, line 1332: Some offtop:

IMHO this point should be redecided, it won't play nicely with switcheable endianness (which means that the struct definition must be exactly the same for be and le). Though enumerating the fields from lower bytes to higher looks more natural, KS interface is from higher bits to lower.

user_guide.adoc
KOLANICH: IMHO just `aaaaabbb bbbbbbcc` was better. What bit is the most significant and which is the least - it is KS convention that MSB goes first within bit-sized type, and it is documented in text.

generalmimon:
@dgelessus 

> I think it would be better if the docs would recommend setting `bit-endian` explicitly for both BE and LE cases.

> The exception is of course when writing specs that need to be compatible with older compiler versions for whatever reason. (...) But I would explain that in the dedicated section on `bit-endian` at the bottom

Fixed in https://github.com/kaitai-io/kaitai_struct_doc/pull/43/commits/080b9eff2083c110a88cbf20acb753ac142d4ba6, I hope it's OK now.

user_guide.adoc
```diff
@@ -1167,25 +1249,190 @@ i.e. two least significant bits of the first byte would be lost and
...
-integers using old-school methods with value instances. Here's the
-very first example with IPv4 packed start, unpacked manually:
+[[bit-ints-le]]
+==== Little-endian order
+
+IMPORTANT: Feature available since v0.9.
+
+Most formats using little-endian _byte order_ with packed multi-byte
+bit fields (e.g. http://formats.kaitai.io/android_img/[android_img],
+http://formats.kaitai.io/rar/[rar] or http://formats.kaitai.io/swf/[swf])
```
dgelessus, line 1259: Minor thing - could you change these URLs to use HTTPS? Or if Asciidoc supports it, it would be even better to use URLs without an explicit protocol, like `//formats.kaitai.io/android_img/`, so that the link uses the same protocol that the current page was accessed with (HTTP or HTTPS).

dgelessus (approved):
Looks good to me, just one minor comment.

user_guide.adoc
generalmimon: OK, makes sense. However, as there are many such links using explicit protocol throughout the docs, I'll find them with fulltext search and fix them all (but outside of this PR), that'll be probably easier.

I still think that in 2020 it totally makes sense to enforce HTTPS on all *.kaitai.io domains (https://github.com/kaitai-io/kaitai_struct/issues/809), but until then, I'll convert the links to use implicit protocol.

generalmimon:
Thanks for your reviews! Merging in.

PR #44, opened 2020-09-27
Preserve http(s) protocol when navigating to *.kaitai.io

generalmimon:
In response to https://github.com/kaitai-io/kaitai_struct_doc/pull/43#discussion_r495605159

Note: `link:` is the syntax for [_link macro_](https://asciidoctor.org/docs/asciidoc-writers-guide/#external-links) in AsciiDoc, it is **not** a special protocol. When you use one of the auto-recognized protocols (`http:` or `https:`), the `link:` macro is not necessary (so instead of `link:https://en.wikipedia.org/[Wikipedia]`, you can write just `https://en.wikipedia.org/[Wikipedia]`), but "no protocol" (`//...`) is not recognized this way, so it has to be done like `link://kaitai.io[Homepage]`.

Cc @dgelessus

### DIFF
--- a/developers.adoc
+++ b/developers.adoc
@@ -249,6 +249,6 @@ After addition, don't forget to update lists of languages:
 * https://github.com/kaitai-io/kaitai_struct — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md — `-t` option documentation
-* https://kaitai.io — everywhere
+* link://kaitai.io[\http://kaitai.io] — everywhere
 * https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view — package description
 * https://twitter.com/kaitai_io — profile
--- a/developers.adoc
+++ b/developers.adoc
@@ -249,6 +249,6 @@ After addition, don't forget to update lists of languages:
 * https://github.com/kaitai-io/kaitai_struct — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md — `-t` option documentation
-* http://kaitai.io — everywhere
+* https://kaitai.io — everywhere
 * https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view — package description
 * https://twitter.com/kaitai_io — profile
--- a/ci.adoc
+++ b/ci.adoc
@@ -17,7 +17,7 @@ not to rely on a single CI engine provider. Currently, we use:
 
 With so many different jobs/flows, it's hard to rely on CI providers
 internal tools (such as job browsers, test browsers, log explorers,
-etc), so we're using our own https://ci.kaitai.io/[CI dashboard] to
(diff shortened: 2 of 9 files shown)


developers.adoc
```diff
@@ -249,6 +249,6 @@ After addition, don't forget to update lists of languages:
 * https://github.com/kaitai-io/kaitai_struct — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler — project description
 * https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md — `-t` option documentation
-* http://kaitai.io — everywhere
+* https://kaitai.io — everywhere
```
dgelessus: Any particular reason why this is explicitly https, unlike all the others?

generalmimon: The reason is "not to confuse users" by showing different URL than the link points to. I just don't want to explicitly show `https://kaitai.io/` when the link actually leads to a http:// version (if we're currently at **http**://doc.kaitai.io/).

Maybe the sensible solution would be to show just `kaitai.io`:

```asciidoc
* link://kaitai.io[kaitai.io] — everywhere
```

but I'm not sure if it looks good among these `https://...` links:

> * https://github.com/kaitai-io/kaitai_struct — project description
> * https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md — `-t` option documentation
> * [kaitai.io](//kaitai.io/) — everywhere
> * https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view — package description
> * https://twitter.com/kaitai_io — profile

Maybe drop the https:// prefixes, as Chrome already does in the address bar?

> * [github.com/kaitai-io/kaitai_struct](https://github.com/kaitai-io/kaitai_struct) — project description
> * [github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/README.md) — `-t` option documentation
> * [kaitai.io](//kaitai.io/) — everywhere
> * [bintray.com/kaitai-io/debian/kaitai-struct-compiler/view](https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view) — package description
> * [twitter.com/kaitai_io](https://twitter.com/kaitai_io) — profile

However, this bloats the markup quite a bit:

```asciidoctor
https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view[bintray.com/kaitai-io/debian/kaitai-struct-compiler/view] — package description
```

vs.

```asciidoctor
https://bintray.com/kaitai-io/debian/kaitai-struct-compiler/view — package description
```

generalmimon: Maybe I'm just bikeshedding here, but as a user I'd be confused if the link that claims to lead on **https**:// version would take me to the **http**:// one.

The opposite is better: many sites are already redirecting automatically from **http**:// to **https**://, so a link that _does the same_ can't be to blame I guess.

Yeah, I'll claim it's `http://kaitai.io/` and will redirect to **http**:// or **https**:// according to the current protocol, because I think I'm just wasting time here.

dgelessus: Ah, I overlooked that here the URL is also the user-visible link text. In that case it's probably fine either way. 🤷

dgelessus (approved):
Thanks for taking care of this change!

PR #45, opened 2020-10-20
User guide fixes

JohnVeness:
I respectfully submit a large number of small changes, mostly grammar, in the user guide.

Please note that I haven't actually used Kaitai Struct yet, so I have tried to mostly stay away from making changes of a technical nature.

I also have a number of questions or things I'm not sure what to do (as with the website repo, it would be good to have Issues available here, but in the absence of that):

1. The doc is inconsistent with regards to using "one" or "you". I would prefer "you", but either way, it should stick with one of them, I feel!
2. Section 4.7 starts "What do we do if we need to use many of the strings in such a format?", which is rather jarring as the strings it is talking about are introduced in section 4.4 I believe, with two unrelated sections after. Maybe 4.4 and 4.7 should be moved next to each other? Or at the very least, a reference to 4.4 should be added to the beginning of 4.7.
3. Section 4.13 introduces calculated (AKA "value") instances, but as far as I can tell, they seem to be called value instances more often in the rest of the doc, and are never actually called calculated instances anywhere else in the doc. Should they maybe just be called value instances throughout?
4. In 4.11, "\`if\`s" is used twice in a bullet point, but it doesn't actually render how you presumably wish it to. I'm not sure what to change it to.
5. In 4.16.1 there is what looks like an unfinished sentence that ends "so keeping just a link to Wikidata is typically enough to."
6. In section 5, there is what looks like an unfinished sentence "Finally, we pass our substream to the Person class instead of"
7. In 7.7 I don't understand the sentence (possibly it is unfinished) "Note that both `OpcodeJmp` and `OpcodePush` supply `this` as `_parent` argument in `Arg` constructor, and, as it is declared as
`KaitaiStruct`."
8. Quite a few places use "KS" as shorthand for "Kaitai Struct". I feel it would be better to use it in long form throughout, but didn't feel bold enough to make such a change.

Many thanks for this document in general. I like its informal style.

Hope this is useful.

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -682,8 +682,8 @@ pattern over and over. Repetition might be:
 
 * element repeated up to the very end of the stream
 * element repeated a pre-defined number of times
-* element repeated while some condition is satisfied (or until some
-condition won't become true)
+* element repeated while some condition is not satisfied (or until some
+condition becomes true)
 
 Kaitai Struct supports all these types of repetitions. In all cases, it will create
 a resizable array (or nearest equivalent available in the target language)
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -418,7 +418,8 @@ dangerous. What if we never encounter that byte?
 
 Another very widespread model is actually having *both* a fixed-sized
 buffer for a string *and* a terminator. This is typically an artifact
-of serializing structures like this from C. For example, if one would take this example structure:
+of serializing structures like this from C. For example, take this
+structure:
 
 [source,c]
 ----
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -29,7 +29,7 @@ chapter will give an overview of the options available.
 
 === Web IDE
 
-If you're going to try Kaitai Struct for the first time, then probably
+If you're going to try Kaitai Struct for the first time, probably
 the Web IDE is the easiest way to get started. Just open
 https://ide.kaitai.io/[Kaitai Struct Web IDE] and you're
 ready to go:
@@ -191,7 +191,7 @@ appending `be` or `le` — i.e. `f4be`, `f8le`, etc
 string has a concept of encoding, which must be specified using
 `encoding`
 
-The YAML-based syntax might look a little more verbose than C-like struct,
+The YAML-based syntax might look a little more verbose than C-like structs,
 but there are a few good reasons to use it. It is consistent, it is
 easily extendable, and it's easy to parse, so it's easy to make your own
 programs/scripts that work with .ksy specs.
@@ -245,7 +245,7 @@ against such blunders.
 
 To specify "magic" bytes (i.e. fixed content) in structures, KS includes
 a special `contents` key. For example, this is the beginning of a `seq`
-for Java .class file:
+for Java .class files:
 
 [source,yaml]
 ----
@@ -262,15 +262,14 @@ is done.
 
 Note that `contents` is very flexible and you can specify:
 
-* An UTF-8 string — bytes from such a string would be used to check
-against
+* A UTF-8 string — bytes from such a string would be checked against
 * An array with:
 ** bytes in decimal representation
 ** bytes in hexadecimal representation, starting with 0x
 ** UTF-8 strings
 
-In case of using an array, all elements' byte representations would be
-concatenated and expected in sequence. Few examples:
+In the case of using an array, all elements' byte representations would be
+concatenated and expected in sequence. Some examples:
 
 [source,yaml]
 ----
@@ -446,7 +445,7 @@ After the first `strcpy` operation, the buffer will look like:
 50 72 69 6e|63 65 73 73|00 ?? ?? ??|?? ?? ?? ??| |Princess.???????|
 ....
 
-And after the second `strcpy`, the the following will remain in the
+And after the second `strcpy`, the following will remain in the
 memory:
 
 ....
@@ -581,13 +580,13 @@ types:
         type: u4
 ----
 
-If `body_len` attribute was in the same type as `body`, we could just
+If the `body_len` attribute was in the same type as `body`, we could just
 use `size: body_len`. However, in this case we've decided to split the
-main header into separate subtype, so we'll have to access it using the `.`
+main header into a separate subtype, so we'll have to access it using the `.`
 operator — i.e. `size: header.body_len`.
 
 One can chain attributes with `.` to dig deeper into type
-hierarchy — i.e. `size: header.subheader_1.subsubheader_1_2.field_4`.
+hierarchy — e.g. `size: header.subheader_1.subsubheader_1_2.field_4`.
 But sometimes we need just the opposite: how do we access upper-level
 elements from lower-level types? KS provides two options here:
 
@@ -606,7 +605,7 @@ TODO
 In some cases, it would be way too impractical to write tons of
 `_parent._parent._parent._parent...` or just plain impossible (if you're
 describing a type which might be used on several different levels, thus
-different number of `_parent` would be needed). In this case, we can use
+different number of `_parent` would be needed). In this case, we can use a
 special pseudo-attribute `_root` to just start navigating from the very
 top-level type:
 
@@ -632,7 +631,7 @@ types:
 
 Some protocols and file formats have optional fields, which only exist
 in some conditions. For example, one can have some byte first that
-designates if some field exists (1) or not (0). In KS, you can do that
+designates if some other field exists (1) or not (0). In KS, you can do that
 using the `if` key:
 
 [source,yaml]
@@ -709,7 +708,9 @@ This yields an array of unsigned integers, each 4 bytes long, which
 spans till the end of the stream. Note that if we've got a number of bytes left in the
 stream that's not divisible by 4 (for example, 7), we'll end up reading
 as much as possible, and the parsing procedure will throw an
-end-of-stream exception. Of course, you can do that with any type,
+end-of-stream exception.
+
+Of course, you can do this with any type,
 including user-defined types (subtypes):
 
 [source,yaml]
@@ -784,8 +785,8 @@ seq:
 This one specifies the `width` and `height` of the matrix first, then parses
 as many `matrix` elements as needed to fill a `width` × `height` matrix
 (although note that it won't be a true 2D matrix: it would still be just
-a regular 1D array, and you'll need to convert (x, y) coordinate to
-address in that 1D array manually).
+a regular 1D array, and you'll need to convert (x, y) coordinates to
+addresses in that 1D array manually).
 
 ==== Repeat until condition is met
 
@@ -832,7 +833,7 @@ types:
 === Typical TLV implementation (switching types on an expression)
 
 "TLV" stands for "type-length-value", and it's a very common staple in
-many formats. The basic idea is that we do a modular and
+many formats. The basic idea is that we use a modular and
 reverse-compatible format. On the top level, it's very simple: we know
 that the whole format is just an array of records (`repeat: eos` or
 `repeat: expr`). Each record starts the same: there is some marker that
@@ -1417,8 +1418,8 @@ types:
 ====
 _Big-endian_ and _little-endian_ bit integers can follow *only on a byte
 boundary*. They can't share the same byte. Joining them on an
-unaligned bit position is _undefined behavior_, and KSC will be throwing
-a compile-time error in the future if it detects such situation.
+unaligned bit position is _undefined behavior_, and future versions of KSC will throw
+a compile-time error if it detects such a situation.
 
 For example, this is *illegal*:
 
@@ -1435,7 +1436,7 @@ seq:
 [[ksy-documentation]]
 === Documenting your spec
 
-We've introduced the `doc` key <<docstrings,early in this user guide>> as
+We introduced the `doc` key <<docstrings,early in this user guide>> as
 a simple way to add docstrings to the attributes. However, it's not
 only attributes that can be documented. The same `doc` key can be used
 in several different contexts:
@@ -1726,7 +1727,7 @@ authors to provide good cross references of their formats:
 
 == Streams and substreams
 
-Imagine that we're dealing with structures of known size. For sake of
+Imagine that we're dealing with a structure of known size. For the sake of
 simplicity, let's say that it's fixed to exactly 20 bytes (but all the
 following is also true if the size is defined by some arbitrarily
 complex expression):
@@ -1762,8 +1763,8 @@ seq:
     size: 20
 ----
 
-Note the subtle difference: we've skipped the `size` in first example
-and added it in the second one. From end-user's perspective, nothing
+Note the subtle difference: we've skipped the `size` in the first example
+and added it in the second one. From the end-user's perspective, nothing
 has changed. You can still access Joe's code and name equally well in
 both cases:
 
@@ -1774,7 +1775,7 @@ r.joe().name() // works
 ----
 
 However, what gets changed under the hood? It turns out that
-specifying `size` actually brings some new features: if you modify
+specifying `size` actually brings some new features: if you modify the
 `person` type to be less than 20 bytes long, it still reserves exactly
 20 bytes for `joe`:
 
@@ -1796,8 +1797,8 @@ types:
         size: 10
 ----
 
-In this example, extra 6 bytes would be just skipped. Alternatively,
-if you'll somehow make `person` to be more than 20 bytes long, it will
+In this example, the extra 6 bytes would just be skipped. Alternatively,
+if you make `person` to be more than 20 bytes long, it will
 trigger an end-of-stream exception:
 
 [source,yaml]
@@ -1818,7 +1819,9 @@ types:
         size: 96
 ----
 
-How does it work? Let's take a look under the hood. Sizeless user type
+How does it work? Let's take a look under the hood.
+
+Sizeless user type
 invocation generates the following parsing code:
 
 [source,java]
@@ -1836,24 +1839,24 @@ KaitaiStream _io__raw_joe = new KaitaiStream(_raw_joe);
 this.joe = new Person(_io__raw_joe, this, _root);
 ----
 
-Every class that KS generates carries a concept of "stream", usually
-available as `_io` member. This is the default stream it reads from
+Every class that KS generates carries a concept of a "stream", usually
+available as an `_io` member. This is the default stream it reads from
 and writes to. This stream works just as you might expect from a
-regular IO stream implementation in you average language: it
-encapsulates reading from files and memory, stores pointer to its
+regular IO stream implementation in a typical language: it
+encapsulates reading from files and memory, stores a pointer to its
 current position, and allows reading/writing of various primitives.
 
-Declaring new user-defined type in the middle of `seq` attributes
-generates new object (usually via constructor call), and this object,
+Declaring a new user-defined type in the middle of the `seq` attributes
+generates a new object (usually via a constructor call), and this object,
 in turn, needs its own IO stream. So, what are our options here?
 
-* In "sizeless" case, we just pass current `_io` along to the new
+* In the "sizeless" case, we just pass the current `_io` along to the new
   object. This "reuses" the existing stream with all its properties:
   current pointer position, size, available bytes, etc.
-* In "sized" case, we know the size a priori and want the object we
+* In the "sized" case, we know the size a priori and want the object we
   created to be limited within that size. So, instead of passing an
-  existing stream, instead we create a new substream that will be
-  shorter and will contain exact number of bytes requested.
+  existing stream, we create a new substream that will be
+  shorter and will contain the exact number of bytes requested.
 
 Implementations vary from language to language, but, for example, in
 Java, the following is done:
@@ -1862,20 +1865,20 @@ Java, the following is done:
 ----
 // First, we read as many bytes as needed from our current IO stream.
 // Note that if we don't even have 20 bytes right now, this will throw
-// an EOS exception on this line, and user type won't even be invoked.
+// an EOS exception on this line, and the user type won't even be invoked.
 this._raw_joe = this._io.readBytes(20);
 
 // Second, we wrap our bytes into a new stream, a substream
 KaitaiStream _io__raw_joe = new KaitaiStream(_raw_joe);
 
-// Finally, we pass our substream to Person class instead of
+// Finally, we pass our substream to the Person class instead of
 this.joe = new Person(_io__raw_joe, this, _root);
 ----
 
-After that, parsing of `person` type will be totally bound to limits
-of that particular substream. Anything in Person class whatsoever
-can't do a thing to original stream — it just doesn't have access to
-that object, period.
+After that, parsing of a `person` type will be totally bound to the limits
+of that particular substream. Nothing in the Person class
+can do a thing to the original stream — it just doesn't have access to
+that object.
 
 Let's check out a few use cases that demonstrate how powerful this
 practice can be.
@@ -1884,9 +1887,9 @@ practice can be.
 
 Quite often binary formats use the following technique:
 
-* First comes some integer that declares total size of the structure
-  (or structure's body, i.e. everything minus this length header).
-* Then comes structure's body, which is expected to have exactly
+* First comes some integer that declares the total size of the structure
+  (or the structure's body, i.e. everything minus this header).
+* Then comes the structure's body, which is expected to have exactly the
   declared number of bytes.
 
 Consider this example:
@@ -1927,8 +1930,8 @@ Thus:
 ----
 
 But this is very inconvenient and potentially error prone. What will
-happen if at some time in future the record contents will be updated
-and we'll forget to update this formula?
+happen if at some time in future the record contents are updated
+and we forget to update this formula?
 
 It turns out that substreams offer a much cleaner solution here. Let's
 separate our "header" and "body" into two distinct user types, and
@@ -1957,15 +1960,15 @@ types:
         size-eos: true
 ----
 
-For `comment`, we just made it to have size up until the end of
+For `comment`, we just made its size to be up until the end of
 stream. Given that we've limited it to the substream in the first
-place, this means exactly what we wanted.
+place, this is exactly what we wanted.
 
 [[repeat-until-size-limit]]
 === Repeating until total size reaches limit
 
 The same technique might be useful for repetitions as well. If you
-have an array of same-type entries, and a format declares total size
+have an array of same-type entries, and a format declares the total size
 of all entries combined, again, you can try to do this:
 
 [source,yaml]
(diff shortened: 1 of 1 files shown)


user_guide.adoc
```diff
@@ -483,15 +482,14 @@ This works in 2 steps:
...
 === Enums (named integer constants)
 
-The nature of binary format encoding dictates us that in many cases
+The nature of binary format encoding dictates that in many cases
 we'll be using some kind of integer constants to encode certain
-entities. For example, a widely known IP packet uses 1-byte integer to
-encode protocol type for the payload: 6 would mean "TCP" (which gives us
-infamous TCP/IP), 17 would mean "UDP" (which yields UDP/IP), and 1 means
+entities. For example, an IP packet uses a 1-byte integer to
+encode protocol type for the payload: 6 would mean "TCP" (which gives us TCP/IP), 17 would mean "UDP" (which yields UDP/IP), and 1 means
```
dgelessus: I think there should be another "the" here:
```suggestion
entities. For example, an IP packet uses a 1-byte integer to
encode the protocol type for the payload: 6 would mean "TCP" (which gives us TCP/IP), 17 would mean "UDP" (which yields UDP/IP), and 1 means
```

user_guide.adoc
```diff
@@ -563,8 +560,8 @@ TODO
 
 === Accessing attributes in other types
 
-Expression language (used, for example, in `size` key) allows you to
-refer not only attributes in current type, but also in other types.
+Expression language (used, for example, in a `size` key) allows you to
+refer not only to attributes in the current type, but also in other types.
```
dgelessus, line 564:
```suggestion
The expression language (used, for example, in a `size` key) allows you to
refer not only to attributes in the current type, but also in other types.
```

user_guide.adoc
```diff
@@ -648,10 +645,10 @@ seq:
 ----
 
 In this example, we again use expression language to specify a boolean
```
dgelessus, line 647:
```suggestion
In this example, we again use the expression language to specify a boolean
```

user_guide.adoc
```diff
@@ -707,11 +704,13 @@ seq:
...
 ----
 
-This yields an array of unsigned integers, each is 4 bytes long, which
+This yields an array of unsigned integers, each 4 bytes long, which
 spans till the end of the stream. Note that if we've got a number of bytes left in the
 stream that's not divisible by 4 (for example, 7), we'll end up reading
-as much as possible, and then parsing procedure will throw an
-end-of-stream exception. Of course, you can do that with any type,
+as much as possible, and the parsing procedure will throw an
+end-of-stream exception.
```
dgelessus:
```suggestion
as much as possible, and then the parsing procedure will throw an
end-of-stream exception.
```

user_guide.adoc
```diff
@@ -1609,14 +1608,14 @@ enums:
 ----
 
 In this format, instead of specifying just the identifier for every
-numeric value, you specify YAML map, which has `id` key for
-identifier, and allows other regular keys (like `doc` and `doc-ref`)
+numeric value, you specify a YAML map, which has an `id` key for
+an identifier, and allows other regular keys (like `doc` and `doc-ref`)
```
dgelessus: I think it would make more sense this way:

```suggestion
numeric value, you specify a YAML map, which has an `id` key for
the identifier, and allows other regular keys (like `doc` and `doc-ref`)
```

user_guide.adoc
```diff
@@ -1668,67 +1667,67 @@ meta:
...
+* `loc` specifies an identifier in the
   https://www.loc.gov/preservation/digital/formats/fdd/browse_list.shtml[Digital
-  Formats] database of https://www.loc.gov/[US Library of Congress], a
+  Formats] database of the https://www.loc.gov/[US Library of Congress], a
   major effort to enumerate and document many file formats for digital
-  preservation purposes. Value typically looks like `fddXXXXXX`, where
+  preservation purposes. The value typically looks like `fddXXXXXX`, where
   `XXXXXX` is a 6-digit identifier.
-* `mime` key is used to specify
+* `mime` specified s
```
dgelessus: Typo?

```suggestion
* `mime` specifies a
```

user_guide.adoc
```diff
@@ -2221,17 +2221,17 @@ types:
...
 
 Some formats obscure the data fully or partially with techniques like
-compression, obfuscation or encryption. In this cases, incoming data
-should be pre-processed before actual parsing would take place, or we'll
-just end up with the garbage getting parsed. All such pre-processing
-algorithms has one thing in common: they're done by some function that
-takes a stream of bytes and return the stream of bytes (note that number
-of incoming and resulting bytes might be different, especially in case
+compression, obfuscation or encryption. In theses cases, incoming data
+should be pre-processed before actual parsing takess place, or we'll
```
dgelessus:
```suggestion
compression, obfuscation or encryption. In these cases, incoming data
should be pre-processed before actual parsing takes place, or we'll
```

user_guide.adoc
```diff
@@ -2334,17 +2334,17 @@ parse flag bitfields, as you can omit `flag_foo != 0` syntax and just
...
 little catch here: the same syntax is used for "true" arrays of
-objects (see below), so if you'll try to do stuff like `[1, 1000, 5]`
+objects (see below), so if you try to do stuff like `[1, 1000, 5]`
 (`1000` obviously won't fit in a byte), you won't get a byte array,
-you'll get array of integers instead.
+you'll get an array of integers instead.
 
 *Strings* normally come from using `type: str` (or `type: strz`, which
-is actually a shortcut that also implicitly adds `terminator: 0`).
+is the shortcut that implicitly adds `terminator: 0`).
```
dgelessus:
```suggestion
is a shortcut that implicitly adds `terminator: 0`).
```

user_guide.adoc
```diff
@@ -2517,8 +2517,8 @@ Can be applied to integers and floats:
 * `a % b` — modulo; note that it's not a remainder: `-5 % 3` is `1`,
   not `-2`; the result is undefined for negative `b`.
 
-NOTE: If both operands are integer, result of arithmetic operation is
-integer, otherwise it is floating point number. For example, that
+NOTE: If both operands are integer, the result of an arithmetic operation is
+integer, otherwise it is a floating point number. For example, that
```
dgelessus: I think using "integer" as an adjective is grammatically correct, but in a programming context I almost always see it used as a noun. I would perhaps write it like this for clarity:

```suggestion
NOTE: If both operands are integers, the result of an arithmetic operation is
an integer, otherwise it is a floating point number. For example, that
```

user_guide.adoc
```diff
@@ -2624,14 +2624,14 @@ literals), and it's possible to call methods on it. The common syntax
...
+`obj.attr_name` syntax. One can chain that to traverse a
 chain of substructures: `obj.foo.bar.baz` (given that `obj` is a
-user-defined type that has `foo` field, which points to user-defined
-type that has `bar` field, and so on).
+user-defined type that has a `foo` field, which points to a user-defined
+type that has a `bar` field, and so on).
 
-There are a few pre-defined methods that form kind of a "standard
+There are a few pre-defined methods that form a kind of "standard
 library" for expression language.
```
dgelessus, line 2635:
```suggestion
library" for the expression language.
```

user_guide.adoc
```diff
@@ -2716,7 +2716,7 @@ library" for expression language.
 
 |`to_i`
 |Integer
-|Converts enum into corresponding integer representation
+|Converts an enum into a corresponding integer representation
```
dgelessus:
```suggestion
|Converts an enum into the corresponding integer representation
```

(every enum value has exactly one corresponding integer value)

user_guide.adoc
```diff
@@ -4010,9 +4011,9 @@ but does not directly specify sizes of the files:
 ....
 
 In this example, we have N = 3 offsets, which specify N - 1 = 2
-sections. Sizes of the sections must be derived as start of the
+sections. Sizes of the sections must be derived as the start of the
 following section minus start of the current section. To do such
```
dgelessus:
```suggestion
following section minus the start of the current section. To do such
```

dgelessus (commented):
Thank you for this big PR! Almost all of these changes look good to me. I've left comments/suggestions in a few places which I think are incorrect or could be worded better, but they are all very minor.

(I'll make a separate comment to try to answer your questions.)

dgelessus:
> 2. Section 4.7 starts "What do we do if we need to use many of the strings in such a format?", which is rather jarring as the strings it is talking about are introduced in section 4.4 I believe, with two unrelated sections after. Maybe 4.4 and 4.7 should be moved next to each other? Or at the very least, a reference to 4.4 should be added to the beginning of 4.7.

Yeah, it looks like the "Substructures" section was supposed to come directly after one of the earlier sections with the length-prefixed string example, and the "Enums" section was later added in between by accident. We could move the section, or reword the introduction a bit so that it reads better even if it doesn't come directly after the string example.

> 3. Section 4.13 introduces calculated (AKA "value") instances, but as far as I can tell, they seem to be called value instances more often in the rest of the doc, and are never actually called calculated instances anywhere else in the doc. Should they maybe just be called value instances throughout?

In my experience, as someone who's mainly a Kaitai Struct *user* and less a compiler developer, they are almost always called "value instance". I think the term "calculated instance" might come from the compiler internals - for example there are many `Calc*Type` classes in the compiler, which I think represent values that come from an expression rather than being read from the stream.

> 4. In 4.11, "\`if\`s" is used twice in a bullet point, but it doesn't actually render how you presumably wish it to. I'm not sure what to change it to.

If I'm reading [the Asciidoctor docs](https://asciidoctor.org/docs/asciidoc-writers-guide/#quoted-text) correctly, the fix for that is to use double backticks, i. e. "\`\`if\`\`s". Single backticks apparently only work if they are surrounded by spaces and not inside a word.

> 8. Quite a few places use "KS" as shorthand for "Kaitai Struct". I feel it would be better to use it in long form throughout, but didn't feel bold enough to make such a change.

Yep, we abbreviate Kaitai Struct as KS quite often, especially on the issue tracker and in the chat. For the docs it might be better to always write Kaitai Struct in full, so that it's clearer for people who are less familiar with the project.

JohnVeness:
Many thanks for your code review suggestions, picking up things I missed and typos I added! I have committed most of them.

Regarding "expression language", I noticed throughout that it was referred to just like that, as if it's a title, without a "the" or a "a" in front of it. I agree that it sounds rather jarring it being referred to that way, but it is so consistent I thought it might be deliberate. I would feel more comfortable if it was capitalised as Expression Language, or called KSEL or something, to make its use as a title feel better. However, as that is such a large change I don't feel bold enough to make that!

JohnVeness:
> Yeah, it looks like the "Substructures" section was supposed to come directly after one of the earlier sections with the length-prefixed string example, and the "Enums" section was later added in between by accident. We could move the section, or reword the introduction a bit so that it reads better even if it doesn't come directly after the string example.

Yeah, I agree. I don't feel bold enough to move sections though, especially if their placement is important (e.g. referenced from elsewhere)!

> In my experience, as someone who's mainly a Kaitai Struct _user_ and less a compiler developer, they are almost always called "value instance". I think the term "calculated instance" might come from the compiler internals - for example there are many `Calc*Type` classes in the compiler, which I think represent values that come from an expression rather than being read from the stream.

Thanks. As this is a *user* guide, I have changed "calculated instance" to "value instance".

> If I'm reading [the Asciidoctor docs](https://asciidoctor.org/docs/asciidoc-writers-guide/#quoted-text) correctly, the fix for that is to use double backticks, i. e. "``if``s". Single backticks apparently only work if they are surrounded by spaces and not inside a word.

Thanks, done.

> Yep, we abbreviate Kaitai Struct as KS quite often, especially on the issue tracker and in the chat. For the docs it might be better to always write Kaitai Struct in full, so that it's clearer for people who are less familiar with the project.

Yep, done.

user_guide.adoc
```diff
@@ -412,15 +412,14 @@ seq:
...
 
 Reading "until the terminator byte is encountered" could be
-dangerous. What if we'll never encounter that byte?
+dangerous. What if we never encounter that byte?
 
 Another very widespread model is actually having *both* a fixed-sized
 buffer for a string *and* a terminator. This is typically an artifact
-of serializing structures like this from C. For example, if one would
-just take this example structure:
+of serializing structures like this from C. For example, if one would take this example structure:
```
generalmimon: Keep the line width:

```suggestion
of serializing structures like this from C. For example, if one would
take this example structure:
```

user_guide.adoc
```diff
@@ -2815,14 +2815,13 @@ a few pre-defined internal methods (they all start with an underscore
 
 ==== Switching over strings
 
-One can use type switching technique over any other comparable values
-beside integers. For example, one can switch over a string value. Note
-that left side (key) of `cases` map is a full-featured KS expression,
+One can use type switching over other comparable values, not just integers. For example, one can switch over a string value. Note
+that the left side (key) of a `cases` map is a full-featured Kaitai Struct expression,
```
generalmimon:
```suggestion
One can use type switching over other comparable values,
not just integers. For example, one can switch over a string value. Note
that the left side (key) of a `cases` map is a full-featured Kaitai Struct expression,
```

user_guide.adoc
```diff
@@ -3023,9 +3022,9 @@ str2 = "bar"
 
 |====
 
-But sometimes you indeed want that dot, and it should be part of
-output. It is actually possible too, if you'll specify `include: true`
-(the default it is false):
+But sometimes you indeed want that dot, and it should be part of the
+output. This is possible, by specifying `include: true`
```
generalmimon: Maybe leave the comma out?

```suggestion
output. This is possible by specifying `include: true`
```

Yeah, this also makes sense with regard to this sentence a few lines below _(NB: no comma)_:

https://github.com/kaitai-io/kaitai_struct_doc/blob/f4ec2562de5bf4f002411e9d0b78b397b7c688a5/user_guide.adoc#L3059-L3060

user_guide.adoc
```diff
@@ -3090,19 +3089,21 @@ the_rest = ".bar."
...
-to "attempted to read past end of stream". However, in some cases, you
-really want to read until the end of stream, and this should be
-considered normal encoding, not an error. For this cases, you can
-specify `eos-error`:
+to "attempted to read past end of stream".
+
+In some cases, you
+really want to read until the end of the stream, and this should be
+considered normal encoding, not an error. For these cases, you can
+specify `eos-error: false` (the default is true):
```
generalmimon: I'd go for

```suggestion
specify `eos-error: false` (the default is `true`):
```

... so that it has the same typeface as the value `false` a few words before.

user_guide.adoc
```diff
@@ -3148,7 +3149,7 @@ seq:
 ----
 
 and you want to use it in a file listing specification
-`filelist.ksy`. Here's how to do that:
+`filelist.ksy`, here's how to do that:
```
generalmimon, line 3153: Good catch, now it makes a lot more sense! But I also suggest to replace _given that_ with _if_ at the beginning of the sentence:

> For example, ~~given that~~ **if** you have a `date.ksy` file that describes the date structure
> _[`date.ksy` contents]_
> and you want to use it in a file listing specification, here's how to do that:

user_guide.adoc
```diff
@@ -3363,8 +3364,8 @@ MyCustomProcessor _process__raw_buf = new MyCustomProcessor(key());
 this.buf = _process__raw_buf.decode(this._raw_buf);
 ----
 
-Typical implementation of a custom processor would look like that
-(again, example for in Java, refer to language-specific notes for
+A typical implementation of a custom processor would look like this
+(again, example in Java; refer to language-specific notes for
```
generalmimon, line 3369: A random thought (I **do not suggest** to address it in this PR at all): I remember that @GreyCat somewhere said it would be nice to show all supported target languages (i.e. C++, C#, Python, ...) for all code samples which show the generated code in one particular target language (mostly Java, currently), using tabs for switching among them.

user_guide.adoc
```diff
@@ -3402,13 +3403,13 @@ processor class should implement. For Java, it's named
   this particular example, it just adds whatever we supplied as "key"
   to every byte
 
-NOTE: `decode` can return different number of bytes that it gets
-(which is perfectly normal, for example, for decompression).
+NOTE: `decode` can return a different number of bytes that it gets
+(which is perfectly normal, for example, with decompression).
```
generalmimon: I don't get this one (maybe it should be like this, but I'm not sure). Is there a rationale behind it, or your feeling?

user_guide.adoc
```diff
@@ -3569,34 +3570,34 @@ types:
...
+recursive structure. This way one can access the structure's root by just
 using `_parent`. Naturally, we've done exactly that to get ourselves
 `chunk_size` by just using `size: _parent.chunk_size`.
 
 ==== Omitting parent
 
-In some cases, you'd rather want some object to don't have any parent
-at all. Primary use case for that is to make sure that some
-instantiation it does not affect parent type. In many cases, resorting
+In some cases, you want some object to not have any parent
```
generalmimon: Shouldn't it be

```suggestion
In some cases, you want some object not to have any parent
```

?

Or maybe

> you want some object to ~~not~~ have ~~any~~ **no** parent at all.

user_guide.adoc
```diff
@@ -19,55 +19,53 @@ fields, repetitions, fields that depend on other fields previously
...
-Kaitai Struct has somewhat diverse infrastructure around it, this
+Kaitai Struct has a somewhat diverse infrastructure around it. This
 chapter will give an overview of the options available.
 
 === Web IDE
 
-If you're going to try Kaitai Struct for the first time, then probably
-it's the easiest way to get started. Just open
+If you're going to try Kaitai Struct for the first time, probably
+the Web IDE is the easiest way to get started. Just open
```
generalmimon:
```suggestion
If you're going to try Kaitai Struct for the first time,
the Web IDE is probably the easiest way to get started. Just open
```

user_guide.adoc
```diff
@@ -19,55 +19,53 @@ fields, repetitions, fields that depend on other fields previously
...
-A list of Web IDE features is available https://github.com/kaitai-io/kaitai_struct_webide/wiki/Features[on _kaitai_struct_webide_ GitHub wiki].
+A list of Web IDE features is available on the https://github.com/kaitai-io/kaitai_struct_webide/wiki/Features[_kaitai_struct_webide_ GitHub wiki].
 
-Note that there are *two* different versions of the Web IDE:
+Note that there are two different versions of the Web IDE:
 
-1. https://ide.kaitai.io/ — *stable* version, has the stable of the Kaitai Struct compiler (currently 0.8, released 2018-02-05)
-2. https://ide.kaitai.io/devel/ — unstable development version, has always the *latest* compiler (the most recent 0.9-SNAPSHOT)
+1. https://ide.kaitai.io/ — *stable* version, has the stable of the Kaitai Struct compiler (currently 0.9, released 2020-10-16)
+2. https://ide.kaitai.io/devel/ — unstable development version, always has the *latest* compiler (the most recent 0.9-SNAPSHOT)
```
generalmimon: Yep, and please also update this:

```suggestion
2. https://ide.kaitai.io/devel/ — unstable development version, always has the *latest* compiler (the most recent 0.10-SNAPSHOT)
```

user_guide.adoc
```diff
@@ -105,14 +103,14 @@ Misc options:
...
 * `--verbose` — verbose output
 * `--help` — display usage information and exit
-* `--version` — output version information and exit
+* `--version` — display version information and exit
 
 == Workflow overview
 
-As you might have already read on the project's website, the main idea
+The main idea
 of Kaitai Struct is that you create a description of a binary data
```
generalmimon:
```suggestion
The main idea of Kaitai Struct is that you create a description of a binary data
```

user_guide.adoc
```diff
@@ -253,23 +254,22 @@ seq:
     contents: [0xca, 0xfe, 0xba, 0xbe]
 ----
 
-This reads first 4 bytes and compares them to 4 bytes `CA FE BA BE`. If
-there would be any mismatch (or there would be less than 4 bytes read),
+This reads the first 4 bytes and compares them to the 4 bytes `CA FE BA BE`. If
+there is any mismatch (or less than 4 bytes are read),
 that'll throw an exception and stop parsing at an early stage, before any
```
generalmimon: Would this be a bit better?

```suggestion
it throws an exception and stops parsing at an early stage, before any
```

user_guide.adoc
```diff
@@ -315,7 +315,7 @@ One of the most common methods used to mitigate this problem is to use
 some integer to designate length of the string, and store only
 designated number of bytes in the stream. Unfortunately, this yields
 variable-length structure, and it's impossible to describe such thing
```
generalmimon:
```suggestion
designated number of bytes in the stream. Unfortunately, this yields
a variable-length structure, and it's impossible to describe such thing
```

user_guide.adoc
```diff
@@ -373,18 +373,18 @@ strings, but the same features should work on any user types as well.
...
 ....
 61 62 63 00
 ....
 
-This 4 bytes actually represent 3-character string "abc", and extra
-one trailing byte "0" (AKA null) serves as a delimiter or
-terminator. By agreement, C strings cannot include 0 byte: every time
+These 4 bytes actually represent the 3-character string "abc", plus one extra
+trailing byte "0" (AKA null) which serves as a delimiter or
+terminator. By agreement, C strings cannot include a 0 byte: every time
```
generalmimon:
```suggestion
terminator. By agreement, C strings cannot include a byte 0: every time
```

user_guide.adoc
```diff
@@ -548,13 +545,13 @@ types:
...
-`types:` key on top-level type. That's a map, inside it we can define as
-many subtypes as we want. We define just one, and inside it we just nest
+just by doing `type: str_with_len`. The type itself is defined using the
+`types:` key at the top level. That's a map, and inside it we can define as
+many subtypes as we want. We define just one, and inside it we nest
 the exact same syntax as we use for the type description on the top
 level — i.e. the same `seq` designation.
 
-NOTE: There's no need for `meta` here, as type name is derived from
+NOTE: There's no need for `meta` here, as the type name is derived from the
```
generalmimon:
```suggestion
NOTE: There's no need for `meta/id` here, as the type name is derived from the
```

user_guide.adoc
```diff
@@ -688,8 +685,8 @@ pattern over and over. Repetition might be:
 * element repeated while some condition is satisfied (or until some
 condition won't become true)
```
generalmimon: To describe the `repeat-until` key, it should be:

```suggestion
* element repeated while some condition is not satisfied (or until some
condition becomes true)
```

user_guide.adoc
```diff
@@ -783,17 +782,17 @@ seq:
...
 ----
 
-This one specifies `width` and `height` of the matrix first, then parses
+This one specifies the `width` and `height` of the matrix first, then parses
 as many `matrix` elements as needed to fill a `width` × `height` matrix
 (although note that it won't be a true 2D matrix: it would still be just
-a regular 1D array, and you'll need to convert (x, y) coordinate to
-address in that 1D array manually).
+a regular 1D array, and you'll need to convert (x, y) coordinates to
+addresses in that 1D array manually).
```
generalmimon: I'd prefer _index_ in the array rather than _address_ in the array:

```suggestion
a regular 1D array, and you'll need to convert (x, y) coordinates to
indices in that 1D array manually).
```

Memory _address_ is not exactly what we're looking for, because that is low level and can be implementation specific. Array index is a sufficiently high level (or _abstract_) and established term 🙂

user_guide.adoc
JohnVeness: Thanks for spotting this. I have resolved this in a different way.

user_guide.adoc
JohnVeness: Done.

user_guide.adoc
JohnVeness: Sounds a good idea.

user_guide.adoc
JohnVeness: I've modified these lines again. Better now?

user_guide.adoc
JohnVeness: Thanks. I've gone for your second suggestion.

user_guide.adoc
JohnVeness: I am unclear on this...

JohnVeness:
Thanks @generalmimon ! I have committed many of your suggestions.

user_guide.adoc
JohnVeness: Not sure about this. I have a feeling (although haven't web searched to check), that people often refer to the "zero byte" in the context of string terminators.

generalmimon: > Not sure about this. I have a feeling (although haven't web searched to check), that people often refer to the "zero byte" in the context of string terminators.

Yeah, you're probably right, this was just my gut feeling, as I have never seen "0 byte" where 0 is written as a numeral. I'd say it's quite unusual. Maybe "zero byte" would be better?

user_guide.adoc
generalmimon: > I am unclear on this...

I don't know how to formulate it clearly, but the attribute is repeated as long as the `repeat-until` condition is `false`, i.e. while the condition is not satisfied. An example can be found below in the User Guide (https://doc.kaitai.io/user_guide.html#_repeat_until_condition_is_met):

```ksy
seq:
  - id: numbers
    type: s4
    repeat: until
    repeat-until: _ == -1
```

And it usually translates to this code (JS):

```javascript
    do {
      var _ = this._io.readS4le();
      this.numbers.push(_);
    } while (!(_ == -1)); // `!` is a negation (NOT) operator
```

And another way how to express "while the condition is not satisfied (i.e. is equal to `false`)" is "until the condition **is** satisfied (i.e. evaluates to `true`)", because the word _until_ in English kind of means "while not ..." as I understand it.

So I think that _"until some condition won't become true"_ doesn't sound too English, because it can be rewritten to _"**while** some condition won't **not** become true"_, and we get a double negative here, which is not usual in English, and it apparently wasn't the intention anyway.

user_guide.adoc
JohnVeness: Yep, I've changed to "zero byte".

user_guide.adoc
JohnVeness: My brain hurts, but I'll accept your suggestion anyway! It could always be improved another time if it needs it.

user_guide.adoc
```diff
@@ -3878,19 +3880,19 @@ specify that, but instead one can use:
 * `bool` to specify booleans
 * `struct` to allow arbitrary KaitaiStruct-compatible user types
 * `io` to specify KaitaiStream-compatible IO streams
-* `any` to allow any type (if target language supports that)
+* `any` to allow any type (if the target language supports that)
 
 NOTE: Parametric types with mandatory parameters can't be read from
-file directly (as they obviously need parameter values supplied from
-somewhere), and thus typical `fromFile(...)` helper method is not
+files directly (as they obviously need parameter values supplied from
```
generalmimon: I would actually prefer the singular form, i.e. _from **a** file_ or _from **the** file_. "Reading from files" might sound like we want to read a single type from several files at once, which is not the case.

JohnVeness: Yep, agreed. Done.

generalmimon (approved):
Huge thanks for this great and valuable contribution, merging in!

PR #46, opened 2020-10-30
Update instructions for publishing releases of the Python runtime

dgelessus:
Some parts of the instructions were a bit outdated - for example they used `setup.py upload` for uploading to PyPI, which has been deprecated for a while and has been removed completely from current setuptools versions. While I was here, I also made some other additions and fixes - see the commit message for a full list.

### DIFF
--- a/developers.adoc
+++ b/developers.adoc
@@ -225,10 +225,8 @@ to `X.Y.Z` format (e.g. `0.9.0` for 0.9 release)
 
 === Python
 
-* Make sure that the needed Python packaging tools are installed and up-to-date:
-** Debian/Ubuntu/etc.: `sudo apt install python3-setuptools python3-wheel twine`
-** Other Linuxes: see https://packaging.python.org/guides/installing-using-linux-tools/[Python Packaging User Guide]
-** Everywhere else: use your system package manager, or `python3 -m pip install --upgrade setuptools twine wheel`
+* Install/update the needed Python packaging tools: `python3 -m pip install --upgrade setuptools twine wheel`
+** If this gives you permission errors, **don't use `sudo` to run `pip`**. Instead use your system package manager (e. g. `apt`) to install/update the tools, or add the `--user` option to `pip install`.
 * Pump version in `kaitaistruct.py`, seek `\\__version__ =`
 * Delete the `dist` directory (if it exists), so that built files from previous or development versions don't get uploaded by accident.
 * `python3 setup.py sdist bdist_wheel`
--- a/developers.adoc
+++ b/developers.adoc
@@ -230,32 +230,15 @@ to `X.Y.Z` format (e.g. `0.9.0` for 0.9 release)
 ** Other Linuxes: see https://packaging.python.org/guides/installing-using-linux-tools/[Python Packaging User Guide]
 ** Everywhere else: use your system package manager, or `python3 -m pip install --upgrade setuptools twine wheel`
 * Pump version in `kaitaistruct.py`, seek `\\__version__ =`
-* Build distributions and upload them to PyPI:
-+
-[source,bash]
-----
-# Delete any old previously built distributions.
-$ rm -r dist/
-# Build source and wheel distributions.
-$ python3 setup.py sdist bdist_wheel
-# Run basic correctness checks on the built distributions.
-$ twine check dist/*
-# Upload the built distributions.
-# You may want to upload to TestPyPI first as a test before uploading to the real PyPI.
-# `twine upload` will prompt for credentials,
-# or you can provide them via the environment variables TWINE_USERNAME and TWINE_PASSWORD
-# (see https://twine.readthedocs.io/en/latest/#environment-variables).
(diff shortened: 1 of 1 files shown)


developers.adoc
```diff
@@ -225,25 +225,37 @@ to `X.Y.Z` format (e.g. `0.9.0` for 0.9 release)
...
-[distutils]
-index-servers =
-    pypi
-
-[pypi]
-repository: https://pypi.python.org/pypi
-username: <login>
-password: <pass>
+# Delete any old previously built distributions.
+$ rm -r dist/
```
generalmimon: Well, _OK_ 😐, but I'd rather leave this out completely. Doesn't it become to some development literacy 😃? I'd say it's somewhat native to clean the generated folders just to be on the safe side, to start with a clean slate before the release. And BTW, have you tested whether it's necessary for the procedure to work?

Yet one more thing: this will fail if the `dist/` doesn't exist, so there should be definitely a `-f` flag:

```suggestion
# Delete any old previously built distributions.
$ rm -rf dist/
```

And the comment is even more obvious and useless I think...

developers.adoc
```diff
@@ -225,25 +225,37 @@ to `X.Y.Z` format (e.g. `0.9.0` for 0.9 release)
...
+$ python3 setup.py sdist bdist_wheel
+# Run basic correctness checks on the built distributions.
+$ twine check dist/*
+# Upload the built distributions.
+# You may want to upload to TestPyPI first as a test before uploading to the real PyPI.
+# `twine upload` will prompt for credentials,
+# or you can provide them via the environment variables TWINE_USERNAME and TWINE_PASSWORD
+# (see https://twine.readthedocs.io/en/latest/#environment-variables).
+$ twine upload --repository=testpypi dist/* # Upload to TestPyPI (https://test.pypi.org/)
+$ twine upload dist/* # Upload to live PyPI (https://pypi.org/)
```
generalmimon: Yeah, I really think that these comments are unnecessary and redundant.

And please keep the style of the document, i.e. not this long shell snippets, but one command per list item, optionally having a short concise comment next to it (but only when it's not clear). Something like this will be enough:

 - `python3 setup.py sdist bdist_wheel`
 - `twine check dist/*`
 - `twine upload dist/*` and fill credentials
   - (`twine upload --repository=testpypi dist/*` to publish to [test.pypi.org](https://test.pypi.org/))

No need to explain what `twine check` or `twine upload` does, I guess the names are self-explanatory 😉 And if someone wants to know more about these commands for whatever reason, all what it takes is to paste them into Google.

developers.adoc
dgelessus: Sure, it's good practice to start with a clean environment when making a release, and if you remember to do that, great. I put this in the instructions as a reminder just in case, so that it's less likely that someone uploads junk files to PyPI by accident.

Especially because with some other languages/tools you don't necessarily have to manually clean up previous build artifacts before making a release. For example when using Gradle to publish a JVM library, it will upload only the jars for the current version (that were just built) to the Maven repo, and not any other random files that might still be under build/libs.

Regarding the comment - yeah, the way I worded it, it's a bit redundant. I suppose if anything I should have explained the problem with accidentally uploading old files by accident.

developers.adoc
dgelessus: Yeah, I reformatted this a few times while writing it. The shell code block made more sense without comments in between 🙂

I commented/explained a bit more than usual, because someone following this guide might be less familiar with how Python packaging works - when you have to publish runtimes for 5 different languages, there will always be some languages that you know less well than others 🙂 But yeah, the commands have relatively descriptive names, so some of these explanations probably aren't necessary.

developers.adoc
```diff
@@ -225,25 +225,20 @@ to `X.Y.Z` format (e.g. `0.9.0` for 0.9 release)
...
-password: <pass>
-----
-* Pump version in `kaitaistruct.py`, seek `__version__ =`
-* `python3 setup.py sdist upload`
-** (use `python3 setup.py sdist upload -r pypitest` to publish to testing server)
-* Check that new version appears at https://pypi.python.org/pypi/kaitaistruct[https://pypi.python.org/pypi/kaitaistruct]`/$VERSION`
+* Make sure that the needed Python packaging tools are installed and up-to-date:
+** Debian/Ubuntu/etc.: `sudo apt install python3-setuptools python3-wheel twine`
+** Other Linuxes: see https://packaging.python.org/guides/installing-using-linux-tools/[Python Packaging User Guide]
+** Everywhere else: use your system package manager, or `python3 -m pip install --upgrade setuptools twine wheel`
```
generalmimon: Doesn't `python3 -m pip install ...` work everywhere? I think there's no need to distinguish operating systems and Linux distributions.

dgelessus: It does, sort of. But when you have Python installed through a package manager (which is common on Linux), it's expected that you install Python packages using the package manager and not `pip`. If you use `pip` anyway, you get at best a confusing situation with multiple versions of the same package installed, or at worst you break the installation managed by the package manager.

I'll rewrite this to be a bit shorter and more general.

generalmimon (approved):
LGTM, thanks!

PR #50, opened 2022-07-07
Modify [Repetition index] example

bjrxyz:
I tink this example maybe have a slip of the pen.


generalmimon (approved):
@bjrxyz Good catch, thanks!

PR #53, opened 2023-07-11
Migrate from Travis CI to GitHub Actions

generalmimon:
See https://github.com/kaitai-io/kaitai_struct/issues/824


.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          echo 'gem: --no-document' >> ~/.gemrc
+          gem install asciidoctor pygments.rb && gem install --prerelease asciidoctor-tabs
+      - name: Install Docson
+        run: git clone -b v2.1.0 --depth 1 https://github.com/lbovet/docson.git
+      - name: Build
+        run: make
+      - name: Deploy
+        env:
+          BOT_SSH_KEY: ${{secrets.BOT_SSH_KEY}}
```
generalmimon, line 35: @GreyCat I've just added `kaitai_struct_doc` to the list of selected repositories that can use `BOT_SSH_KEY` in GitHub Actions, so hopefully this migration should go smoothly:

image

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
```
generalmimon, line 17: Perhaps it would be good to upgrade to Ruby 3 (since Ruby 2 is EOL according to https://endoflife.date/ruby and https://www.ruby-lang.org/en/downloads/branches/), but the purpose of this PR is just to migrate from Travis without much additional changes and the Travis pipeline [was running on Ruby 2.7](https://app.travis-ci.com/github/kaitai-io/kaitai_struct_doc/jobs/598856275#L269):

```console
$ ruby --version
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]
```

After we see GitHub Actions working, we can look into upgrading to Ruby 3.

.github/workflows/main.yml
GreyCat: Sounds good! Thanks so much for this vital piece of infrastructure!

.github/workflows/main.yml
GreyCat: Yep, makes sense, let's do that as a separate update.

generalmimon:
@GreyCat I've also converted `gem install` invocations to a `Gemfile` in the subsequent commits (https://github.com/kaitai-io/kaitai_struct_doc/compare/66a2b23...64d3580). It seems to be a more standard way of installing dependencies in Ruby and it also appears to have reduced the execution time of the CI pipeline (the [job using `gem install`](https://github.com/kaitai-io/kaitai_struct_doc/actions/runs/5544279112/jobs/10121504366) took 1m 39s, the [job using `Gemfile`](https://github.com/kaitai-io/kaitai_struct_doc/actions/runs/5545261254/jobs/10123928452) took 53s - the approach with `Gemfile` is probably faster due to [caching](https://github.com/ruby/setup-ruby/tree/v1.152.0#caching-bundle-install-automatically) provided by the [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby) action).

On the other hand, I found that `bundle install` doesn't make executable scripts from installed gems globally available on PATH, so now we have to run it with `bundle exec`. But perhaps it's a good thing to choose a installation approach that doesn't affect & isn't affected by the global environment (not really a concern in CI, but might be on a local machine).

PR #54, opened 2023-12-26
Add "Validating attribute values" section to user_doc

mblsha:
Adapted the discussion thread [1] to document the expected `valid` key behavior.

[1]: https://github.com/kaitai-io/kaitai_struct/issues/435

### DIFF
--- a/user_guide.adoc
+++ b/user_guide.adoc
@@ -310,6 +310,47 @@ definitely not recommended in real-life specs):
 NOTE: There's no need to specify `type` or `size` for fixed contents
 data — it all comes naturally from the `contents`.
 
+[[valid-values]]
+=== Validating attribute values
+
+To ensure attributes in your data structures adhere to expected formats and ranges, Kaitai Struct provides a mechanism for validating attribute values using the `valid` key. This key allows you to define constraints for values, enhancing the robustness of your specifications. Here's how you can enforce these constraints:
+
+* `eq` (or directly `valid: value`): Ensures the attribute value exactly matches the given value.
+* `min`: Specifies the minimum valid value for the attribute.
+* `max`: Specifies the maximum valid value for the attribute.
+* `any-of`: Defines a list of acceptable values, one of which the attribute must match.
+* `expr`: An expression that evaluates to true for the attribute to be considered valid.
+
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks a lot for starting this one! I will add some more, but it's a good clean start.

user_guide.adoc
```diff
@@ -310,6 +310,47 @@ definitely not recommended in real-life specs):
...
+
+  - id: calculated_value
+    type: u4
+    valid:
+      expr: _ % 2 == 0  # Value must be even
+----
+
+When a value does not meet the specified criteria, Kaitai Struct raises a validation error, halting further parsing. This preemptive measure ensures the data being parsed is within the expected domain, providing a first layer of error handling.
+
+NOTE: The actual implementation of validation checks is language-dependent and may vary in behavior and supported features across different target languages.
```
generalmimon, line 352: I don't understand why this note is here - it goes against the principles of Kaitai Struct. Once everything is correctly implemented in all target languages (which may with any KS feature temporarily not be the case, but it's simply the reality that individual target languages are in various stages of development at any point in time), the checks will not vary in behavior across different target languages. The status (passed/failed) of any check in one language must be consistent with the status of the same check in the same situation in any other language, any violation is a bug that will be eventually fixed.

@mblsha Did you have something specific in mind when you wrote this sentence? I've just checked the `Valid*` tests at https://ci.kaitai.io/ and I don't see any current issues with `valid`, except for Nim, Perl and Construct (all of which are somewhat niche targets), where `valid` presumably isn't implemented at all, but I don't see any behavior variations.

I remember that `valid` still doesn't have type checks (https://github.com/kaitai-io/kaitai_struct_formats/pull/678#discussion_r1261783881), so perhaps you mean the implications of that.

GreyCat: I believe what @mblsha meant here was that exception handling by itself is not a universal mechanism by any stretch, and indeed individual exception names and notation will vary widely between languages. That's at least how I interpreted this. I certainly won't argue that we should also have language-dependent explanations on how these validation exceptions (or error codes, or whatever it will manifest in) work for specific languages.

mblsha:
@generalmimon should I delete this "NOTE:" from the doc then?

